PR #3, opened 2012-05-22
New BeanHas and BeanProperty matchers

sandromancuso:
The idea is to be able to test multiple properties in a single assertion and get a decent error message when at least one of them does not match. 

I've written these matchers and we use them in a few projects at work. We thought that it would be a good idea to contribute back to JavaHamcrest. I changed my original classes to so it would fit better the coding style used in JavaHamcrest, including the way the tests are written. 

I know that Hamcrest has the allOf and hasProperty matchers. They are very good but I personally think that BeanHas and BeanProperty could be a good addition to the library since they work in a slightly different way.

For more details, I've blogged about the matchers here: http://craftedsw.blogspot.co.uk/2012/05/testing-multiple-properties-with-single.html

### DIFF
--- a/hamcrest-library/src/main/java/org/hamcrest/beans/BeanHas.java
+++ b/hamcrest-library/src/main/java/org/hamcrest/beans/BeanHas.java
@@ -12,8 +12,8 @@ import org.hamcrest.StringDescription;
  * <p>How to use it: <p>
  * 
  * // Static imports <br/>
- * import static org.craftedsw.beanpropertymatcher.matcher.BeanMatcher.has; <br/>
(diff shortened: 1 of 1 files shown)


mageddo:
Oh god, im am looking for that @scarytom  @sandromancuso

rpyaset:
> I'm curious -- what does this add that can't be done by combining multiple hasProperty into an allOf?

compile time safety

nhojpatrick:
@sandromancuso fancy rebasing this from master, as `hamcrest-core` and `hamcrest-library` have been refactored a lot and also deprecated, so that everything is just in `hamcrest`.

nhojpatrick:

PR #9, opened 2012-09-20
Add method describeMismatch to IsNot matcher

lanwen:
This is useful thing that i use in my everyday work. Without this, my own matchers with not(), cant describe their mismatch

### DIFF
--- a/hamcrest-core/src/main/java/org/hamcrest/core/IsNot.java
+++ b/hamcrest-core/src/main/java/org/hamcrest/core/IsNot.java
@@ -30,14 +30,19 @@ public class IsNot<T> extends BaseMatcher<T>  {
         description.appendText("not ").appendDescriptionOf(matcher);
     }
 
-    
+
+    @Override
+    public void describeMismatch(Object item, Description description) {
+        matcher.describeMismatch(item, description);
+    }
+
     /**
      * Creates a matcher that wraps an existing matcher, but inverts the logic by which
      * it will match.
      * <p/>
      * For example:
      * <pre>assertThat(cheese, is(not(equalTo(smelly))))</pre>
-     * 
+     *
      * @param matcher
      *     the matcher whose sense should be inverted
      */
@@ -53,7 +58,7 @@ public class IsNot<T> extends BaseMatcher<T>  {
      * <pre>assertThat(cheese, is(not(smelly)))</pre>
      * instead of:
      * <pre>assertThat(cheese, is(not(equalTo(smelly))))</pre>
-     * 
+     *
      * @param value
      *     the value that any examined object should <b>not</b> equal
      */


andreas-lundgren-storytel:
I am not experienced enough to put in any strong arguments, but let me describe the issue I had in #165 in more detail.

I have a test case using a custom Matcher. The test case looks like this:
```
    @Test
    public void testRoutingConfiguration() {
        // This test case tests the production configuration of the routes.
        // It can be used to assure that specific paths requires specific scopes in production.
        assertThat(getRouting("/auth/login"), is(exposedToExternalWeb()));
        assertThat(getRouting("/auth/logout"), is(not(exposedToExternalWeb())));
        assertThat(getRouting("/auth/logout/swagger.yaml"), is(exposedToExternalWeb()));
        assertThat(getRouting("/auth/update_token"), is(not(exposedToExternalWeb())));
...
```

My matcher looks like this:
```
import com.example.api.routing.RouteService;
import org.hamcrest.Description;
import org.hamcrest.Factory;
import org.hamcrest.Matcher;
import org.hamcrest.TypeSafeMatcher;

import java.util.Optional;

public class ExposedToExternalWeb extends TypeSafeMatcher<Optional<RouteService.RouteConfig>> {

    @Override
    protected boolean matchesSafely(Optional<RouteService.RouteConfig> config) {
        if (config.isPresent()) {
            return !config.get().isFilter();
        } else {
            return false;
        }
    }

    @Override
    public void describeTo(Description description) {
        description.appendText("exposed to external web");
    }

    /**
     * Creates a matcher that matches if the examined {@link RouteService.RouteConfig} is exposed to the external web
     * <p/>
     * For example:
     * <pre>assertThat(getRouting("some url path"), is(exposedToExternalWeb())</pre>
     * <pre>assertThat(getRouting("some other url path"), is(not(exposedToExternalWeb()))</pre>
     */
    @Factory
    public static Matcher<Optional<RouteService.RouteConfig>> exposedToExternalWeb() {
        return new ExposedToExternalWeb();
    }

    @Override
    protected void describeMismatchSafely(Optional<RouteService.RouteConfig> item, Description mismatchDescription) {
        Optional optional = (Optional) item;
        if (!optional.isPresent()) {
            mismatchDescription.appendText("tested path was not described in the yml-configuration.");
        } else {
            RouteService.RouteConfig config = (RouteService.RouteConfig) optional.get();
            mismatchDescription.appendText("path \"")
                .appendText(config.getRoute())
                .appendText("\" was ")
                .appendText("described with filter parameter: ")
                .appendText(String.valueOf(config.isFilter()));
        }
    }
}
```

With the current implementation of not, if a negatet testcase fails (in this case the assertion for "/auth/logout"), I get the error (using IntelliJ):
```
java.lang.AssertionError: 
Expected: is not exposed to external web
     but: was <Optional[com.storytel.api.routing.RouteService$RouteConfig@2f75a9b1]>
Expected :is not exposed to external web
     
Actual   :<Optional[com.storytel.api.routing.RouteService$RouteConfig@2f75a9b1]>
 <Click to see difference>
```

If I override org.hamcrest.core.IsNot and add the method:
```
    @Override
    public void describeMismatch(Object item, Description mismatchDescription) {
        matcher.describeMismatch(item, mismatchDescription);
    }
```
I get a valid output looking like:
```
java.lang.AssertionError: 
Expected: is not exposed to external web
     but: path "/auth/logout" was described with filter parameter: false
```

PR #16, opened 2012-12-12
Issue 100 (Google Code):  hasItem matcher factory has wrong generic wildcard

qualidafial:
Currently, the Matchers.hasItem(s) factory methods accept one or more
Matcher<? super T> and return a Matcher< Iterable<? super T>>. The wildcard
in the return type makes it difficult to use hasItem matchers in assert
statements.

This commit modifies the factory method return type to simply
Matcher< Iterable<T\>\>.


modax:
@qualidafial I don't think so. As far as I can tell, it is not possible to get:

```java
assertThat(List<? extends CharSequence>, hasItem(Matcher<CharSequence>))
```

working with the current hamcrest version (2.0.0.0). Could anything  be done about this as the problem was report long ago?

PR #26, opened 2013-01-18
Expose paramater type info where available to help with Java 7 issues.

louisth:
Lots of matchers know what type of object they are dealing with, but there's currently no way to get at that information. This type information would be really helpful for some Hamcrest clients, such as JMock, in dealing with the new typing issues introduced in Java 7. (Specifically, with this fix JMock clients won't have to change their tests when JMock is made Java 7 compilable.) I propose adding a getParameterType method to Matcher that lets the matchers that know their type expose that information (and other Matchers can just return null). BaseMatcher provides the default implementation so clients do not have to make any code changes.

I'm bundling two other small fixes. AnyOf and AllOf now have the same base class, and the mismatch description for TypeSafeDiagnosingMatcher now matches TypeSafeMatcher. If this is a problem, I'd be happy to resubmit these separately.


nhojpatrick:
@louisth fancy rebasing this from master, as `hamcrest-core` and `hamcrest-library` have been refactored a lot and also deprecated, so that everything is just in `hamcrest`.

nhojpatrick:

PR #28, opened 2013-02-06
Improve describeMismatch output across the board.

louisth:
After running into confusing output from describeMismatch and reading the discussion at hamcrest/JavaHamcrest#9, I have tackled the problem head on. I believe @sf105 and @ihaworth have the right idea. Instead of trying to describe the failure, we need to describe the given item from the Matcher's point of view. According to this contract for describeMismatch, the existing implementations are basically correct.

What I've done is three things.
1. I extended and revised the Javadoc to specify this new contract for describeMismatch:

```
    /**
     * Generates a description of the given item from the Matcher's point of view.
     * The description will be part of a larger description of why a matching
     * failed, so it should be concise.
     * The description should be able to replace Y in the sentence "Expected X but Y," and
     * should be in the past tense, for example, "Expected null but <U>was &lt;"foo"&gt;</U>."
     * The description should NOT describe the matcher, but rather should highlight features of interest on the item as they actually are.
     * The description must ALWAYS be filled in, regardless of return value of {@link #matches}. 
      */
```

(DescribeMismatch is now rather misnamed, but I leave it to the maintainers to decide whether they want to break backwards compatibility to fix this.)
1. I wrote DescribeMismatchTortureTest, which calls describeMismatch on EVERY matcher in Hamcrest. I used not() with every matcher. I used a variety of matcher compositions including anyOf() and allOf(), and sub-matchers with a variety of descriptions, such as equalTo(), instanceOf(), and startsWith(). I matched against a variety of target objects. I'm sure there are interesting cases I missed but it's a start. What's nice is the torture test allows us to see if a description is phrased right to fit in a variety of contexts. It also allows us to check for reasonable consistency across all the matchers.
2. I fixed every broken description and mismatch description until every test in the torture test produced a reasonable result. Improvements are welcome, but I think this is a good start.

For anyone disagrees with my solution and wants to try their own hand at tackling this problem, I encourage you to at least grab the torture test. I think it does a pretty good job of illustrating and exercising the complexity of the problem.

### DIFF
--- a/hamcrest-library/src/main/java/org/hamcrest/io/FileMatchers.java
+++ b/hamcrest-library/src/main/java/org/hamcrest/io/FileMatchers.java
@@ -12,19 +12,28 @@ import org.hamcrest.TypeSafeDiagnosingMatcher;
 
 public class FileMatchers {
 
+    private static void describeBasically(File file, Description mismatchDescription) {
+        if (!file.exists()) {
+            mismatchDescription.appendText("was nonexistent");
+        } else if (file.isDirectory()) {
+            mismatchDescription.appendText("was a directory");
+        } else if (file.isFile()) {
+            mismatchDescription.appendText("was a file");
+        } else {
+            mismatchDescription.appendText("was special (not a file or a directory)");
+        }
+    }
+    
     @Factory
     public static Matcher<File> anExistingDirectory() {
         return new TypeSafeDiagnosingMatcher<File>() {
             public boolean matchesSafely(File actual, Description mismatchDescription) {
-                final boolean result = actual.isDirectory();
-                if (!result) {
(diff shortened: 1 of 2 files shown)


nhojpatrick:
@louisth  fancy rebasing this from master, as `hamcrest-core` and `hamcrest-library` have been refactored a lot and also deprecated, so that everything is just in `hamcrest`.

sf105:
Personally, I think this one will cause unexpected problems. The reason we kept it slightly clunky is that getting rid of the clumsiness takes a lot of code and is never quite right. The uses cases here are straightfoward, but it gets messy very quickly. Then you're left with a lot more complicated code and still the same problem. I strongly recommend not taking this one, or at least simplifying it.

PR #29, opened 2013-02-22
Improve generics on some collection matchers. 

fbarber:
The following classes are no longer parameterized as they don't care what is in
the collection they match:
- IsCollectionWithSize
- IsIterableWithSize
- IsMapWithSize
- IsEmptyCollection
- IsEmptyIterable

### DIFF
--- a/hamcrest/src/main/java/org/hamcrest/Matchers.java
+++ b/hamcrest/src/main/java/org/hamcrest/Matchers.java
@@ -739,7 +739,7 @@ public class Matchers {
    * @param sizeMatcher
    *     a matcher for the length of an examined array
    */
-  public static <E> org.hamcrest.Matcher<E[]> arrayWithSize(org.hamcrest.Matcher<? super java.lang.Integer> sizeMatcher) {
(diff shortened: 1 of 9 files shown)


guw:
@sf105 Are those methods available already?

I'm running into generics compilation errors with the JDT 3.20 compiler for the following code:
```java
List<String> result = ...
assertThat(result, allOf(hasSize(1), hasItems("foo")));
```

The error is:
_The method allOf(Matcher<? super T>, Matcher<? super T>) in the type Matchers is not applicable for the arguments (Matcher<Collection<? extends Object>>, Matcher<Iterable<String>>)_

nhojpatrick:
@fbarber fancy rebasing this from master, as `hamcrest-core` and `hamcrest-library` have been refactored a lot and also deprecated, so that everything is just in `hamcrest`.

I've got some custom collections matchers, which do similar and would be interested getting richer collection matches into hamcrest. i.e. if the collection is the wrong size then output the collection, so when debugging you can shortcut having to run it yourself and get a start on working out the issue.

fbarber:
Sure.  I will update with a rebased version soon.

fbarber:
@nhojpatrick Rebase done.

nhojpatrick:

Issue #11, opened 2012-09-27
Hamcrest should be an osgi bundle


rgrunber:
I've seen discussion of this at http://code.google.com/p/hamcrest/issues/detail?id=173 , and the topic has come up in regards to doing the same for Junit (https://github.com/KentBeck/junit/issues/212). Is there any plan for this ?

cvgaviao:
Any news about this issue?

josephw:
@cvgaviao This is addressed by #98 . If you're an interested OSGi user, please comment over there to confirm that it's a fix.

PR #32, opened 2013-04-07
OSGi Bundle for issue #11

Tibor17:
An update of  the build script in order to provide OSGi Headers in JavaHamcrest.


sf105:
Hoping that this is fixed with https://github.com/hamcrest/JavaHamcrest/pull/98

PR #38, opened 2013-07-19
generic static factory methods for IsAnything matcher

innokenty:


sf105:
Can you write up an example where this would be useful? Thx

innokenty:
So it just doesn't complain about unsafe generics..

sf105:
Just tried it in a test (Java 7) and not seeing any problems. Can you help?

innokenty:
I don't remember already :smiley: 
It's like if you write
```java
String s = "";
assertThat(s, is(anything()));
```
IDEA should complain about generics mismatch. But you can just close it if it doesn't...

sf105:
Works for me. Please reopen with an example if I'm wrong.

PR #45, opened 2014-01-02
ExtractMatcher (Second-order matcher)

joao-de-melo:
More information: http://jmelo.lyncode.com/hamcrest-second-order-matcher-a-missing-matcher/


markhobson:
I've released [Hamcrest Compose](https://github.com/markhobson/hamcrest-compose) which provides a `hasFeature` matcher along these lines.

PR #46, opened 2014-02-25
Add a cleanup method to handle wrong conversion of type like java.util.Map.Entry

boretti:
Outside of the code format, this change is related to the method :

private static String typeToString(Type type). The returned string, for the Type case, is passed thru a new cleanup method that correct wrong output with parameter of the type :
java.util.Map.Entry<K,V> (without the fix, produce java.util.Map.java.util.Map$Entry<K,V>) which is not valid in java.

The file use a regex to do the correction :
```java
private static String cleanup(String input) {
  return input.replaceFirst("^(.+)\\.\\1\\$", "$1.");
}
```


sf105:
I've ported the regex matcher. Unfortunately, we've dropped the whole Factory thing now in preparation for a restructuring of the Matchers class.

Issue #49, opened 2014-03-13
TypeSafeDiagnosingMatcher/CombinableMatcher conflicts with nullValue Matcher


mmichaelis:
The following assertion never evaluate to true:

``` java
assertThat(null, either(equalTo(something)).or(nullValue()))
```

The root cause lies in the CombinableMatcher being a TypeSafeDiagnosingMatcher which implements `matches()` as:

``` java
public final boolean matches(Object item) {
    return item != null
        && expectedType.isInstance(item)
        && matchesSafely((T) item, new Description.NullDescription());
}
```

Thus for `null` the matchers will never be queried. I suppose the solution should be something like:

``` java
public final boolean matches(Object item) {
    return (item == null || expectedType.isInstance(item))
        && matchesSafely((T) item, new Description.NullDescription());
}
```

StarBax89:
This is caused by this little pice of code in TypeSafeDiagnosingMatcher
```
@Override
    @SuppressWarnings("unchecked")
    public final boolean matches(Object item) {
        return item != null
            && expectedType.isInstance(item)
            && matchesSafely((T) item, new Description.NullDescription());
    }
```
The `item != null`  and `expectedType.isInstance(item)` is false for `item = null`

This can be fixed by adding this dirty hack to CombinableMatcher
```
@Override
  public final boolean matches(Object item) {
    if(this.matcher instanceof AnyOf)
    {
      AnyOf anyOf = (AnyOf) this.matcher;
      List<Matcher<T>> matchers = (List<Matcher<T>>) anyOf.getMatchers();
      Boolean matches = false;
      for(Matcher eachMatcher : matchers)
      {
        if(eachMatcher.matches(item))
        {
          matches = true;
        }
      }
      return matches;
    }
    else {
      return super.matches(item);
    }
  }
```
Therefore TypeSafeDiagnosingMatcher's matches method must be non final and ShortcutCombination needs a getter for matchers.

I know this is really dirty, maybe someone can make a fix with this information. I'll also try to write a better fix for that.

EDIT:
I also wrote a test
`private static final CombinableMatcher<String> EMPTY_STRING_OR_NULL = CombinableMatcher.either(equalTo("")).or(nullValue());`
   
```
@Test public void
    acceptNullOrEmptyString() {
        assertMatches("or didn't pass", EMPTY_STRING_OR_NULL, null);
        assertMatches("either didn't pass", EMPTY_STRING_OR_NULL, "");
    }
```

PR #56, opened 2014-08-18
Proposed solution to Issue #49 TypeSafeDiagnosingMatcher/CombinableMatcher conflicts with nullValue Matcher

hughwphamill:
Proposing a solution to this issue by adding a static factory method which wraps and AllOf with IsNull as the first entry and input Matcher as the second.


sf105:
Thanks for your support. We already have an IsEmptyString matcher, which I think is the most common case. If you find another common use case, perhaps we can revisit.

PR #57, opened 2014-08-23
Add a matcher to check a string against a regular expression

### LINKED ISSUE #36: Matcher checking a string matches a regular expression
I cannot find any class allowing me to write my tests like:

```
String actual = tested.buildUrlWithRandomPort("foo", "bar");
assertThat(actual, matchesRegex("^foo://bar:\\d+/$");
// fails with a message like:
// expected <foo://bar:/> to match a regular expression <^foo://bar:\d+/$>
```

I think that Hamcrest Library should offer matchesReges() together with containsString(), endsWith() and startsWith().

Is there a matcher like this available in the master branch? If not, I could write one (something similar to [Hamcrest Regex Matcher](http://piotrga.wordpress.com/2009/03/27/hamcrest-regex-matcher/)) and create a pull request.

josephw:
This resolves #36

### DIFF
--- a/hamcrest-library/src/test/java/org/hamcrest/object/MatchesPatternTest.java
+++ b/hamcrest-library/src/test/java/org/hamcrest/object/MatchesPatternTest.java
@@ -28,7 +28,7 @@ public class MatchesPatternTest {
     }
 
     @Test
-    public void doesNotMatchesDifferentString() {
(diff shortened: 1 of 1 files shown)


everflux:
This seems to be part of the 2.0.0.0 release which is available since January.

PR #60, opened 2014-09-13
Add a Matcher which allows the use of lamdas to test the field of an object

wjam:
The idea behind this matcher is to reduce the need to write custom matchers under Java 8 by being able to use a lambda to retrieve the field inside an object to match against


markhobson:
I've released [Hamcrest Compose](https://github.com/markhobson/hamcrest-compose) which provides a `hasFeature` matcher along these lines.

npryce:
Since this is alreadt implemented in auxialiary libraries and duplicate's the FeatureMatcher class, we will drop this for now and revisit when we move over to Java 8 entirely.

PR #79, opened 2014-12-04
New matcher HasSubsequence

bhamiltoncx:
When implementing a library to test compiler command-line arguments in the form:

`-x foo -y -z bar`

I found I needed a Hamcrest matcher which would look for a consecutive
sequence of items anywhere inside a collection of items.

I was surprised that Hamcrest had matchers which:

1) Match if a collection includes a single item anywhere (`hasItem`)
2) Match if a collection includes multiple items anywhere (`hasItems`)
3) Match if a collection includes multiple items in order anywhere, consecutive or not (`containsInRelativeOrder`)
4) Match if a collection exactly matches another collection in order (`contains`)
5) Match if a collection exactly matches another collection in any order (`containsInAnyOrder`)

but not for a consecutive sequence of items anywhere inside the collection.

This implements such a matcher (`hasSubsequence`) and adds tests. I borrowed the tests from `containsInRelativeOrder`,
so let me know if there are better tests I should include. I wasn't sure about the `WithValue` stuff
so I just copy-pasted it from that test.

### DIFF
--- a/hamcrest-library/matchers.xml
+++ b/hamcrest-library/matchers.xml
@@ -19,7 +19,7 @@
     <factory class="org.hamcrest.core.StringEndsWith"/>
 
     <!-- Collection -->
-    <factory class="org.hamcrest.collection.HasConsecutiveItems"/>
+    <factory class="org.hamcrest.collection.HasSubsequence"/>
     <factory class="org.hamcrest.collection.IsArray"/>
     <factory class="org.hamcrest.collection.IsArrayContaining"/>
     <factory class="org.hamcrest.collection.IsArrayContainingInOrder"/>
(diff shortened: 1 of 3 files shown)


bgertzfield:
Just back from the holidays now. Thanks for the merge!

I agree that `Collection` can optionally support `add()` and other mutation methods, but since those are optional methods, you're totally right that using it in an API basically means you can't rely on any optional methods.

It's useful, then, to simply treat `Collection` as `Iterable` plus `size()`. In this case, the extra copy to an `ArrayList` isn't going to really kill performance, so it's fine to hide the `Collection` interface if we don't want it.

PR #94, opened 2015-01-18
Improve Every mismatch descriptions (Issue 91)

declension:
* Every now diagnoses *all* mismatches, in a (fairly) friendly way.
 * Extra test case added to illustrate this
 * Self description tweaked to be slightly more consistent with others.
 * Uses a newly added SelfDescribingText class.

### DIFF
--- a/hamcrest-core/src/main/java/org/hamcrest/core/Every.java
+++ b/hamcrest-core/src/main/java/org/hamcrest/core/Every.java
@@ -1,9 +1,10 @@
 package org.hamcrest.core;
 
-import org.hamcrest.Description;
-import org.hamcrest.Factory;
-import org.hamcrest.Matcher;
-import org.hamcrest.TypeSafeDiagnosingMatcher;
+import org.hamcrest.*;
+import org.hamcrest.internal.SelfDescribingText;
+
+import java.util.ArrayList;
+import java.util.List;
 
 public class Every<T> extends TypeSafeDiagnosingMatcher<Iterable<? extends T>> {
     private final Matcher<? super T> matcher;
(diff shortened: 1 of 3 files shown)


hamcrest-core/src/main/java/org/hamcrest/core/Every.java
```diff
@@ -14,19 +15,41 @@ public Every(Matcher<? super T> matcher) {
...
+    private SelfDescribingText mismatchString(T t) {
+        Description desc = new StringDescription();
+        matcher.describeMismatch(t, desc);
+        return new SelfDescribingText(desc.toString());
     }
 
     @Override
     public void describeTo(Description description) {
-        description.appendText("every item is ").appendDescriptionOf(matcher);
+        description.appendText("a collection in which every item is ").appendDescriptionOf(matcher);
```
UrsMetz: As the `Every` matcher can also be used for `Iterables`: should the text talk about "iterables"?

UrsMetz:
@declension I think the description should be more generic (see my line comment). I also find the tests (now) a bit difficult to read as the functionality become more involved: looking at the top with the matching involving substring matching and down again to the test case one might find it hard to now what the result of the test should be. Maybe a submatcher like `equalTo` would be this easier. (note: I'm not a maintainer, these are just my personal thoughts ;-)).

declension:
Yes, for the naming I kinda think the same,  but went for consistency in the end (with `IsCollectionContaining` particularly, which has pretty much the same type signatures).

I'll see if that test can be made a bit more readable / cleaner then.

UrsMetz:
Good point @declension that there are other matchers for `Iterables` that describe themselves using "collection". It surely reads more nicely but is also a bit misleading.

sf105:
The mismatch messages were intended to be used in context. Something like (off the top of my head),
` Expected a collection with every item less than 2 but: 
  mismatches were ['was 4', 'was 6']`
Also, we have to be careful about listing all the mismatches, because sometimes the target collection can be large and swamp the output. I'll take another look, but I think a rejection.

PR #96, opened 2015-01-24
Rewrite HasPropertyWithValue to be a FeatureMatcher

tomwhoiscontrary:
HasPropertyWithValue and FeatureMatcher look really similar to me, and it was driving me nuts that they're completely unrelated. So i made them related.

I have no idea if this is something you would actually want to merge, but i thought someone might at least find the idea mildly interesting.

Because FeatureMatcher has its own ideas about how mismatch descriptions look, i have changed the messages expected in the tests for HasPropertyWithValue. I think the new messages look at least as good as the old ones.

One source of complication in this was that FeatureMatcher::featureValueOf has no way to fail cleanly, so i introduced an [Either](http://book.realworldhaskell.org/read/error-handling.html#errors.either)-like type to wrap the extracted feature value (or its absence). This whole thing would be briefer and more idiomatic-for-Java if FeatureMatcher::featureValueOf was able to throw some sort of exception to signal that a feature's value could not be obtained. FeatureMatcher::matchesSafely could catch this and use its message as a mismatch description. This change would not be strictly backwards-compatible, but the only code which would suffer would be that calling FeatureMatcher::featureValueOf, which would be an odd thing to do.

### DIFF
--- a/hamcrest-library/src/main/java/org/hamcrest/beans/HasPropertyWithValue.java
+++ b/hamcrest-library/src/main/java/org/hamcrest/beans/HasPropertyWithValue.java
@@ -6,6 +6,7 @@ import org.hamcrest.Matcher;
 import org.hamcrest.TypeSafeDiagnosingMatcher;
 
 import java.beans.PropertyDescriptor;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 
 import static org.hamcrest.Condition.matched;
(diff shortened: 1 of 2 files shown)


tomwhoiscontrary:
I've taken this commit off my master so i can do other stuff. It's still around, on a different branch:

https://github.com/tomwhoiscontrary/JavaHamcrest/tree/pr-Rewrite-HasPropertyWithValue-to-be-a-FeatureMatcher

PR #99, opened 2015-03-27
HasPropertyWithValue reports actual exception thrown by bean method

gjoseph:
When readMethod.invoke() was throwing an InvocationTargetException (i.e when the readMethod's execution was throwing an exception), the actual exception and message were lost (e.getMessage() of the InvocationTargetException is null).
With this patch, we report the actual exception's type and message.


sf105:
Accepted with alterations. Thx

PR #102, opened 2015-05-12
Add just in time message creation for the reason

ashleyfrieze:
Following a discussion in JUnit - https://github.com/junit-team/junit/issues/1133 - this is a potential change to allow for the reason message to be evaluated just in time.

Feedback appreciated.


npryce:
If this is for integration with JUnit, please apply it to hamcrest-junit.

npryce:
We already have the SelfDescribing interface, and a method on Description add the description of a SelfDescribing instance to the description's stream.  So instead of defining a new interface, I think it would be better to use the SelfDescribing interface for the lazily calculated message.

ashleyfrieze:
The aim of this pull request is to allow for something like this

    MatcherAssert.assertThat(() -> "Checking validity of " + someBigObject.toString(), 
        someBigObject.isValid());

or

    MatcherAssert.assertThat(() -> "Random access to " + someBigObject.toString(), 
        someBigObject.getThing(), is(42));

In the above cases, let's assume that the standard matchers (or boolean) are good enough to explain the differences, but that the "reason" is something which is non trivial to calculate - e.g. a toString which has to work on a lot of properties. Rather than creating some big strings on the fly when the tests are passing, this would allow those strings to be provided as a lambda or as the implementation of a method.

@npryce If replacing JustInTimeMessage with SelfDescribing would be a hamcresty way of achieving that, then that's great (and easy to change).

I guess a better question is whether the whole change is worth making at all. As I mentioned it came from the discussion over at junit. What do folks think overall?

npryce:
If using SelfDescribing, your example with lambdas would be:

        MatcherAssert.assertThat((d) -> d.appendText("Checking validity of ").appendValue(someBigObject), 
        someBigObject.isValid());

Which would format the message consistently with the rest of the diagnostics.  The Description passed to the lambda would make it easier to format lists, arrays, iterables etc.

ashleyfrieze:
@npryce - yeah, I've been pondering this and I think it's worth a shot, though it's a slightly more long-winded syntax, it's also a richer output.

I'll commit against this pull request and let's see how the code looks. I didn't like the idea of having duplicate implementations inside the assertThat's, but I've hit a small issue with using SelfDescriber to implement the String versions of assertThat for backwards compatibility. The SelfDescribingValue<T> class causes the description to enwrap the reason with " ". I'll come up with a workaround, though open to any suggestions on how best to achieve this.

ashleyfrieze:
As recommended by @npryce, this is now implemented against the SelfDescribing interface. Feedback would be appreciated.

hamcrest-core/src/main/java/org/hamcrest/MatcherAssert.java
```diff
@@ -1,16 +1,19 @@
...
 public class MatcherAssert {
     public static <T> void assertThat(T actual, Matcher<? super T> matcher) {
         assertThat("", actual, matcher);
     }
     
     public static <T> void assertThat(String reason, T actual, Matcher<? super T> matcher) {
+    	assertThat(selfDescribingString(reason), actual, matcher);
+    }
+    
+    public static <T> void assertThat(SelfDescribing reason, T actual, Matcher<? super T> matcher) {
```
kcooney: I would suggest adding some Javadoc that shows how this would be used with Lambdas.

I'm surprised that there isn't any Javadocs in this class at all :-(

ashleyfrieze:
Added Javadoc as per @kcooney 's feedback.

ashleyfrieze:
I'm abandoning this as it's clearly not going to make anyone's day... ever ;)

PR #103, opened 2015-05-18
add Matcher#and, a fluent way to build AllOf

alb-i986:
Example of usage:

```
assertThat(alphabet, startsWith("a").and(endsWith("z"));
```

If you like the idea I will also add Matcher#or.
Then should we deprecate CombinableMatcher?


npryce:
We've experimented with this several times but have never found a satisfactory design w.r.t. precedence and binding of 'and' and 'or' pseudo-operators compared to the built-in logical operators. Java doesn't have operator overloading, so a LISP-y style always feels more natural.

Adding methods to Matcher is a big deal with a large client base (people do ignore the BaseMatcher class despite the weird warning method).  So any syntactic sugar should be implemented separately from the core interfaces.

PR #104, opened 2015-05-18
add BaseMatcher#equals

alb-i986:
This implementation is quite naive and relies on toString:

```
return toString().equals(obj.toString());
```

Considering that toString in turn relies on Description, which is pretty accurate and usually includes all of the variables (actual and expected), it should be actually quite good despite naive.


npryce:
"Usually" isn't a strong enough guarantee for equals().  We haven't specified equals to be part of the Matcher contract, so there are lots of existing implementations of Matcher that will not work and will break the equals/hashCode contract (in fact this implementation breaks the equals/hashCode contract because it does not implement hashCode).

PR #109, opened 2015-06-17
Waiting Matcher

mmichaelis:
This pull request contains a matcher implementing a wait pattern as introduced in [Haste makes waste][haste-minds]. The matcher will wait for the matched object to reach a certain state within a given timeout.

As standard matchers won't check a mutable aspect of the matched object this PR contains in addition a matcher which retrieves an aspect of the matched object and hands it over to some standard Hamcrest matcher.

Thus a typical usage looks like this:

```java
ComponentUnderTest cut = ...;
State expectedState = ...;
Function<ComponentUnderTest,State> stateFunction = new Function<>() { ... };
assertThat(componentUnderTest,
           waitFor(applying(stateFunction, equalTo(expectedState)), 2, SECONDS));
```

In Java 8 this one will look even better - and would not require the preliminary function interface which comes with the `ApplyingMatcher` contained in this PR.

We use the given pattern since 2012 now with great success especially in UI tests where it takes some time until a wanted state is reached.

The wait-algorithm is decelerating regarding polls to the component under test. This is a learning when we first tried to poll very frequently that the component under test was busy answering the polls instead of trying to reach the wanted state (in this case we were polling Solr for having indexed a given document).

[haste-minds]: <http://minds.coremedia.com/2012/11/29/haste-makes-waste/> "Haste makes waste | Minds"

### DIFF
--- a/hamcrest-library/src/main/java/org/hamcrest/Matchers.java
+++ b/hamcrest-library/src/main/java/org/hamcrest/Matchers.java
@@ -1617,7 +1617,7 @@ public class Matchers {
    * @param timeoutMs       timeout in milliseconds
    * @param <T>             accepted value type of the original matcher
    * @return matcher which waits
-   * @see #stateMatches(org.hamcrest.concurrent.WaitFor.WaitForFunction, Matcher)
+   * @see #applying(org.hamcrest.function.Function, Matcher)
    * @see #waitFor(Matcher, long, java.util.concurrent.TimeUnit)
    */
   public static <T> Matcher<T> waitFor(Matcher<T> originalMatcher, long timeoutMs) {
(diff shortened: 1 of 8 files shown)


npryce:
This is useful functionality but this implementation goes against the convention that matching is a pure function.

There are other Hamcrest extension libraries that perform polling & synchronisation in different ways (e.g. see http://github.com/sf105/proboscis). 

Therefore I think this should be published as a third-party library that extends Hamcrest and not be part of the core.

When you publish this as a library, let us know the URL and we'll add it to the page of Hamcrest Extensions.

mmichaelis:
Thanks. The matcher is meanwhile available as [Hamcrest &mdash; Next Deed](https://github.com/mmichaelis/hamcrest-nextdeed) library (already available at Maven Central). Hope you like the name and can guess what it means :smile:

npryce:
Heh. Extended. Good one!

--Nat

PR #110, opened 2015-06-19
Bumped Gradle, added IDE setup to build

szczepiq:
1. Bumped Gradle version to 2.4 to pick up perf improvements and other improvements. Tested by running the build. 2.4 release is much better than 2.3, especially in terms of performance.
2. Added ide setup to the build script files so that developer can invoke "./gradlew idea" to generate IDEA metadata and smoothly import the project into the IDE. Tested by running './gradlew idea' and then opening the generated project in IDEA.


sf105:
Upgraded Gradle to 2.4. I prefer to use gradle import from IDEA rather than generating the project. Let the Idea people decide what's best...

PR #118, opened 2015-10-12
Add coverage reporting

Shredder121:
This at least gives some insight to the lines covered.
I wouldn't recommend enforcing a minimum level of coverage, but I use it to see what I looked at so that I know what to look at next.

If you use your own API, you know what to alter.

_EDITED_:
For reference, [my README.md](https://github.com/Shredder121/JavaHamcrest/tree/6f754ecdbefa5a03c16b50c53052ae9034939780#readme) or alternatively [my README.md using the svg version](https://github.com/Shredder121/JavaHamcrest/tree/4ffee70af25d028ef64925bc032100b0ad9411af#readme).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,6 @@ jdk:
 
 script:
   - ./gradlew clean build javadoc
+
+after_success:
+- ./gradlew jacocoTestReport coveralls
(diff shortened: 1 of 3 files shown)


Shredder121:
Note: I used [this configuration](https://github.com/kt3k/coveralls-gradle-plugin#use-with-jacoco-plugin), so if it's not really the way you configure your build let me know.

sf105:
Thanks for your contribution. I'm very keen to keep dependencies to a minimum so I prefer not to add a coverage tool. If you run coverage and find a hole in the tests, please post that as an issue.

Shredder121:
Ah okay, that makes sense.
Thanks for looking at this though!
Keep up the good work! :+1:

PR #120, opened 2015-10-14
Fix SubstringMatcher's

alb-i986:
fix javadocs
fix NPE in case substring is null: throw IllegalArgumentException instead


sf105:
I've reimplemented. Thanks

PR #123, opened 2015-10-31
Map entry matchers

tomwhoiscontrary:
This adds two complementary matchers for working with maps.

The main matcher is `IsMapWithEntries`, which is just a `FeatureMatcher` for Map::entrySet. This is essentially an escape hatch from the world of maps to the world of true collections; it allows a user to write matches on maps in terms of their set of entries, using the rich set of matchers available for collections (principally `contains` and friends).

The second matcher is `IsMapEntry`, which makes it easy to write matches on the map entries themselves.

I end up writing these two matchers, in some form, in most projects i work on, because this approach so useful. Specifically, i think it's much more useful to be able to view maps as collections than it is to have a range of matchers specifically for maps.

It's possible i'm simply insane, and nobody else will find these useful. But just in case they do, here they are!

### DIFF
--- a/hamcrest-library/src/main/java/org/hamcrest/Matchers.java
+++ b/hamcrest-library/src/main/java/org/hamcrest/Matchers.java
@@ -1,7 +1,5 @@
 package org.hamcrest;
 
-import org.hamcrest.collection.IsCollectionContaining;
-
 import java.util.Collection;
 import java.util.List;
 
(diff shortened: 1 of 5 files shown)


sf105:
I can't get this to merge javadoc is broken because it doesn't recognise @Deprecated, rather than @deprecated. Please fix.

tomwhoiscontrary:
@sf105 Is that comment meant for this PR? I'm on my phone so can't be sure, but I don't think there's any @deprecated, of any capitalisation, in this PR.

sf105:
You branched from an old version of hamcrest that doesn't build in Java 8 any more. I might just branch again and fix it.

S

tomwhoiscontrary:
Got it. I'll jump on this at the weekend, when I'm back in front of a computer.

tomwhoiscontrary:
Okay, so:

1. I based my branch off the v2.0.0.0 branch, for some reason
1. My branch merges cleanly into the v2.0.0.0 branch, and you've done that
1. The v2.0.0.0 branch prior to my commits did not build under Java 8 due to javadoc problems, and is still broken after my commits
1. The master branch is not a descendant of the v2.0.0.0 branch
1. The master branch does build under Java 8
1. Neither my branch nor the old v2.0.0.0 branch merge cleanly into the master branch

So whilst the problems on the v2.0.0.0 branch are not of my making, the problem that my branch is built on that branch is!

It seems like the best thing would be to merge v2.0.0.0 into master, as you have done in the past; however, i'm not sure i know enough about the changes in either branch to do that properly. Alternatively, i could cherry-pick my changes onto master, and submit a new PR. Or i could just leave it to you. What would you like me to do?

tomwhoiscontrary:
This has now become #160.

PR #125, opened 2015-11-10
Escape backslashes properly in BaseDescription

philandstuff:
Before this commit, the following assertion:

```java
assertThat("\n", equalTo("\\n"));
```

gives the output:

```
java.lang.AssertionError:
Expected: "\n"
     but: was "\n"
```

This is confusing because it seems like two identical strings have not
compared equal.

The problem is that in BaseDescription.toJavaSyntax(char) escapes
newline characters but doesn't escape backslash characters, so the two
strings being compared are transformed to the same output for display,
despite having different actual content.

I haven't written a test for this as I couldn't find a test for any of the existing escaping behaviour.


sf105:
Done.  There is a BaseDescriptionTest that you could have extended.

PR #127, opened 2015-11-28
Fix for #27 "The signature of isA is completely wrong".

### LINKED ISSUE #27: The signature of isA is completely wrong
public static <T> org.hamcrest.Matcher<T> isA(java.lang.Class<T> type) {
    return org.hamcrest.core.Is.<T>isA(type);
  }

means you have to cast the thing to the type you are asserting it should be. Really it should be a matcher of anything (Object) not T.

e.g. 
class hierarchy Foo extends Bar

Bar thing = somemethod();
assertThat(thing, isA(Foo.class))

to assert that you are getting the right subclass back - yeah ok, it might not be the best test, but anyhow......

bitzl:
The method `isA(...)` is documented as shortcut to `is(instanceOf(...))`. This was not the case, as the enhanced test `IsA.providesConvenientShortcutForIsInstanceOf()` does not compile (the test was created in analogy to the tests of `instanceof(...)`).

To fix this, the created Matchers need to accept any class instead of the one to check against.

Fixes #27 "The signature of isA is completely wrong".

### DIFF
--- a/hamcrest-core/src/main/java/org/hamcrest/core/Is.java
+++ b/hamcrest-core/src/main/java/org/hamcrest/core/Is.java
@@ -69,7 +69,7 @@ public class Is<T> extends BaseMatcher<T> {
      * <pre>assertThat(cheese, is(instanceOf(Cheddar.class)))</pre>
      * 
      */
-    public static <T> Matcher<T> isA(Class<T> type) {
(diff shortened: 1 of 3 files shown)


sf105:
Implemented.

PR #128, opened 2015-12-04
Adding IsTraversedAs matcher

erosb:
Hello,

this `IsTraversedAs` matcher implementation checks that if an expected and an actual `Iterable`
* return the same number of elements
* return equal elements

I would firstly find it useful, for example when my SUT returns an `ArrayList` instance and I create my expected list using an `Arrays.asList()` call (that is something I find myself doing quite often).

An other usecase can be developing a self-written `Iterable` implementation (the developer may find the `IsTraversedAs` matcher useful for asserting).

Please review it. Thanks.


sf105:
how does this differ from IsIterableContainingInOrder.contains() ?

erosb:
Yes, that seems to be the same (although its name is not good at all).

PR #136, opened 2016-02-25
Boolean getter support added (Issue #138)

buraequete:
This request is to support the `is` prefix for Boolean object getters. Since PropertyDescriptor that is created lacks `readMethod` if the getter does not start with `get`. For example, classes that are generated by JAXB, always contain getters for Boolean types with `is` prefix, and without this minor fix, I cannot match those fields in any of my UT cases.


bureaquete:
Will there be any response to this pull request? Is there any possibility for this fix to be included in `hamcrest:master` ? I do not understand why Boolean getters with `is` prefix are being ignored?

mikulucky:
@bureaquete This is probably not being included because your PR has failed the Continuous Integration

bureaquete:
@mikulucky why though? Everything checks out in my local, no test failure or anything. Cannot find out the reason of CI failure.

mikulucky:
@bureaquete the test failures seem to be with your changes on OpenJDK7 https://travis-ci.org/hamcrest/JavaHamcrest/builds/111651291

bureaquete:
@mikulucky I haven't changed anything but the file in my commits. If you check last couple pull requests, all have the same failure, but from console I cannot figure out the exact reason.

zaphod42:
It looks like the build is hitting a problem in Travis combined with a problem in the JDK. See https://github.com/travis-ci/travis-ci/issues/5227

This isn't anything to do with these specific changes.

bureaquete:
@zaphod42 I tried to use the similar fix from your link, but did not resolve the build failure. Weird issue indeed...

sf105:
I've just added a test for what I think this problem is and it works on my machine without any code changes (Java 7). Please submit a failing test.

sf105:
The fundamental problem appears to be that the JAXB people didn't read the JavaBean spec. That said, I just tried a version of your code and couldn't get it to work because there's no property descriptor for is<BooleanObjectProperty>. Please add a test to demonstrate the behaviour.

bureaquete:
@sf105 I couldn't reproduce the issue I have been facing... In my code, somehow I can get a `PropertyDescriptor` for the `Boolean` field with an empty getter method, and my first modification was to fix that issue. 

But in Hamcrest tests I couldn't replicate the case above, I tried to use exact same class from my project, but still got null `PropertyDescriptor`. Regardless I have added an extra code in `PropertyUtil` to cover this empty case.

nhojpatrick:
@buraequete fancy rebasing this from master, as `hamcrest-core` and `hamcrest-library` have been refactored a lot and also deprecated, so that everything is just in `hamcrest`.

PR #151, opened 2016-06-23
Replace String to CharSequence in text-related matchers.

grumblerchester:


grumblerchester:
Hi
This is the first attempt to update the string-related matchers in order to support interoperability between String, StringBuffer, StringBuilder, CharBuffer, etc (according to found comment).
If code is ok and is useful I gonna update javadocs and increase test coverage.
Please provide your suggestions.

PR #154, opened 2016-07-09
MultiAssert

kypvalanx:
added an asserter that can make multiple asserts regardless of failure and fails at the end providing output at that time.

### DIFF
--- a/hamcrest-core/src/main/java/org/hamcrest/MatcherMultiAssert.java
+++ b/hamcrest-core/src/main/java/org/hamcrest/MatcherMultiAssert.java
@@ -1,0 +1,70 @@
+package org.hamcrest;
+
+import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


nhojpatrick:
@kypvalanx junit jupiter how has assertAll does provider a better solution? And I think TestNG has SoftAsserts or something like that which is similar.

My personal view is if x, y and z are being tested, from your solution if x is false then y and z are not checked, I might have read the code wrong. But using assertAll you can check x and y and z and show them all instead of fixing x, then having to rerun only to see y is also broken.

If you do want to try and get it merged, can you rebase this from master, as `hamcrest-core` and `hamcrest-library` have been refactored a lot and also deprecated, so that everything is just in `hamcrest`.

kypvalanx:
i forgot this was open.  doesn't seem worth it to update hamcrest when there's a newer version

PR #173, opened 2017-02-22
Feature/string contains in any order

tpraxl:
I kindly request to add `stringContainsInAnyOrder` as a complement to `stringContainsInOrder`. It matches if all of the substrings are contained within the subject, regardless of the order.

I'm aware that you are able to achieve the same with

```
assertThat("this is a string", allOf(
  contains("string"),
  contains("this"),
  contains("a")));
```

But there are cases when you need to work with an unknown list of strings, like String[] or List<String>.

For these cases, it would be nice to have

```
String[] tokens = classUnderTest.getTokens();
String subject = getTemplate();
assertThat(subject, stringContainsInAnyOrder(tokens))
```

### DIFF
--- a/hamcrest-library/src/main/java/org/hamcrest/Matchers.java
+++ b/hamcrest-library/src/main/java/org/hamcrest/Matchers.java
@@ -1508,6 +1508,34 @@ public class Matchers {
   }
 
   /**
+   * Creates a matcher of {@link String} that matches when the examined string contains all of
+   * the specified substrings, regardless of the order of their appearance.
+   * For example:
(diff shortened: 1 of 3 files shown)


yankee42:
Just as another workaround for now, if you are on Java8:
```java
assertThat(
    subject,
    allOf(Stream.of(tokens).map(Matchers::containsString).collect(Collectors.toList()))
);
```

nhojpatrick:
@tpraxl looks a good idea, happy to look at getting merged if rebased from master as `hamcrest-core` and `hamcrest-library` has been merged into `hamcrest` so can't currently merge in at the moment.

nhojpatrick:

PR #175, opened 2017-04-28
Added 'hasEntry' matchers for (key, valueMatcher) and (keyMatcher, value)

Shpota:
When I test maps I often need to apply a matcher for the value and check equality for the key. Currently there is no `hasEntry` matcher that consumes a key, and a matcher for the value so I have to do it somewhat like this:
```java
Map<String, List<Integer>> map = singletonMap("key", asList(1, 2));

assertThat(map, hasEntry(equalTo("key"), contains(1, 2)));
```
It would be more convenient to do it like this:
```java
assertThat(map, hasEntry("key", contains(1, 2)));
```

For that reason, I introduced two more `hasEntry` matchers with signatures:
```java
hasEntry(K key, Matcher<? super V> valueMatcher)
```
```java
hasEntry(Matcher<? super K> keyMatcher, V value)
```

### DIFF
--- a/hamcrest-library/src/main/java/org/hamcrest/Matchers.java
+++ b/hamcrest-library/src/main/java/org/hamcrest/Matchers.java
@@ -1114,6 +1114,38 @@ public class Matchers {
 
   /**
    * Creates a matcher for {@link java.util.Map}s matching when the examined {@link java.util.Map} contains
+   * at least one entry whose key equals the specified <code>key</code> <b>and</b> whose
+   * value satisfies the specified <code>valueMatcher</code>.
+   * For example:
+   * <pre>assertThat(myMap, hasEntry("bar", equalTo("foo")))</pre>
+   *
+   * @param key
+   *     the key that, in combination with the valueMatcher, must be describe at least one entry
+   * @param valueMatcher
+   *     the value matcher that, in combination with the key, must be satisfied by at least one entry
+   */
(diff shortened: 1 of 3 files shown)


kvr000:
+1 on this, especially the first one is useful as you usually know exact key and want to have some vague match for the value.

Also, simplifies the unit test runtime complexity from linear to constant (assuming the map is usually HashMap), And some Maps don't support iterating through entries.

Shpota:
@kvr000 this PR has been hanging for 3 years with no response from the maintainers. I have a serious doubt that it will ever be accepted.

nhojpatrick:
@Shpota does look a good idea, can you rebase from master as `hamcrest-core` and `hamcrest-library` has been merged into `hamcrest` so can't currently merge in currently.

sf105:
One caveat is to make sure the overloading works with the extra combinations. Might be worth just doing `hasEntry(K key, Matcher<? super V> valueMatcher)`

Shpota:
@nhojpatrick the original fork is already gone, that's why I cannot rebase it. Please take a look at [this new PR](https://github.com/hamcrest/JavaHamcrest/pull/310) where I applied exactly the same changes on top of the latest version.

@sf105 the overloading does work. There are also tests for that. 

I would also prefer keeping both matchers. At least from my use cases, both are applicable, although the `hasEntry(Matcher<? super K> keyMatcher, V value)` is applicable not that often.

Shpota:
@nhojpatrick could you please take a look at [the updated PR](https://github.com/hamcrest/JavaHamcrest/pull/310)?

nhojpatrick:

PR #178, opened 2017-07-16
Allow Iterables in AllOf/AnyOf that are typed to an extension of `Mat……cher`

yankee42:
Currently the following code won't compile:
```java
List<Matcher<String>> submatchers = new ArrayList<>();
new AllOf<String>(submatchers);
```
This is because `AllOf` expects an `Iterable<Matcher<? super String>>` and not a `Iterable<Matcher<String>>` the I tried to supply above.

As a workaround this works:
```java
List<Matcher<? super String>> submatchers = new ArrayList<>();
new AllOf<String>(submatchers);
```
But depending on where this list comes from it may be impracticable to change the type like that.

This also works, but generates a compile warning and right so...:
```java
List<Matcher<String>> submatchers = new ArrayList<>();
new AllOf<String>((Iterable)submatchers);
````

This PR proposes a slight change in the method signature of `AllOf`, `AnyOf` and `ShortcutCombination` which eliminates the problem. The change in signature is compatible to the old API, because only the generics are changed.


nhojpatrick:
@yankee42 wanting to kick start getting pr's getting reviewed and then merged, can you rebase from master, as `hamcrest-core` and `hamcrest-library` have been refactored a lot and also deprecated, so that everything is just in `hamcrest`.

yankee42:
@nhojpatrick : I rebased it.

nhojpatrick:

Issue #90, opened 2014-12-26
Add String length matcher


helsing:
Are there other than me that would like to see a simple String length matcher in the core library for the upcoming release?

It was requested already more than a year ago, here https://code.google.com/p/hamcrest/issues/detail?id=191

(Please, let me know if I can help out in any way?)

mle-enso:
Let me know, if you agree with this PR or if something is missing.

agebhar1:
would be awesome if this PR could be merged :smile:

derTobsch:
Just googled it :) because I missed this matcher.

sf105:
It's in.

PR #179, opened 2017-07-24
Implementation for CharSequence length matcher (issue #90)

mle-enso:


krichter722:
IDE-specific files should be ignored in `.git/info/exclude` rather than `.gitignore` in order to keep the repository IDE-agnostic.

mle-enso:
Et voilà, now they are where they belong. Happy PR merging. :-)

krichter722:
@agebhar1 It was just a 2 cents contribution. However, @mle-enso the newline at the end of `.gitignore` should be removed as well.

krichter722:
You should provide the changes in one commit, see https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git for an explanation how to do that.

agebhar1:
Thank you both @krichter722 and @mle-enso.

sf105:
Thanks for this. I have a couple of suggestions.
- move to hamcrest-library org.hamcrest.text
- take a look at some of the other size/length matchers, you'll see that they're based on FeatureMatcher and that they also include a matcher of the length.
- I also suggest changing `length` to `hasLength` to make the test read better.

sf105:
I made the changes anyway.

PR #186, opened 2017-10-11
185 Java9 Automatic-Module-Name

nhojpatrick:
as per ticket 185


tumbarumba:
Thanks for the contribution, @nhojpatrick!

PR #200, opened 2018-04-01
Changed signature of Matcher.matches(Object) to matches(T)

### LINKED ISSUE #199: Matcher.matches needs to have parameterized argument
Generics are hard to understand and a wonderful thing after you did. Roughly speaking, the description

```
public interface Matcher<T> extends SelfDescribing {

    /**
     * Evaluates the matcher for argument <var>item</var>.
     *
     * This method matches against Object, instead of the generic type T. This is
     * because the caller of the Matcher does not know at runtime what the type is
     * (because of type erasure with Java generics). It is down to the implementations
     * to check the correct type.
     *
     * @param actual the object against which the matcher is evaluated.
     * @return <code>true</code> if <var>item</var> matches, otherwise <code>false</code>.
     *
     * @see BaseMatcher
     */
    boolean matches(Object actual);
```

shows that generics have not been understood by the author.

> This is because the caller of the Matcher does not know at runtime what the type is (because of type erasure with Java generics)

The opposite (almost) is the case. If you specify a `Matcher` (without generics) as static type the matcher doesn't know the type at runtime - how could it. If you specify a `Matcher<SomeType>` you either

  1. are accessing it at least once as `Matcher` and might get `ClassCastException`s, but also an `unchecked` conversion or `rawtypes` warning which you shouldn't take easy in Java - roughly speaking you get what you deserve or
  2. you use `Matcher<SomeType>` everywhere you don't get warning or compilation errors the Matcher knows _exactly_ which type the argument is at runtime.

`2.` allows to write awesome cast- and type-check-free code in Java and `1.` allows you to do what you want by omitting the generic type on static type of callers, but still be able to enforce maximum type safety in interfaces and thus allow others to stick with `2.`.

It makes sense that the parameter is not used if you don't use it in `matches`.

Once this is fixed the type-safety could be expanded to factory methods like `public static org.hamcrest.Matcher<java.lang.Object> anything()` in `Matchers` which could be `public static <T> org.hamcrest.Matcher<T> anything()` invoking

```
public static <T> Matcher<T> anything(String description) {
    return new IsAnything<>(description);
}
```

Since getting a `Matcher<Object>` if you could have a `Matcher<T>` with no harm done to any other user is slightly painful.

I'm willing to submit a PR if you ok it.

krichter722:
…parameter of the type.

The method comment indicated that this design decision was the result of a misunderstanding of how Java Generics work. The change increases type-safety while users preferring a type-unsafe approach can still omit the generic from the static type of their declarations.

Unused tests which no longer make sense have been removed.
    
WIP: bean-related tests fail and it's very hard to figure out how to pass matchers down the bean graph which formerly had the wrong type (probably returned because they where TypeSafeMatchers)

Closes #199

### DIFF
--- a/hamcrest-core/src/main/java/org/hamcrest/BaseMatcher.java
+++ b/hamcrest-core/src/main/java/org/hamcrest/BaseMatcher.java
@@ -17,7 +17,7 @@ public abstract class BaseMatcher<T> implements Matcher<T> {
     }
 
     @Override
-    public void describeMismatch(Object item, Description description) {
+    public void describeMismatch(T item, Description description) {
         description.appendText("was ").appendValue(item);
     }
 
--- a/hamcrest-core/src/main/java/org/hamcrest/CoreMatchers.java
+++ b/hamcrest-core/src/main/java/org/hamcrest/CoreMatchers.java
@@ -75,7 +75,7 @@ public class CoreMatchers {
    * @param values
    *     optional values to insert into the tokenised description
    */
-  public static <T> org.hamcrest.Matcher<T> describedAs(java.lang.String description, org.hamcrest.Matcher<T> matcher, java.lang.Object... values) {
+  public static <T> org.hamcrest.Matcher<T> describedAs(java.lang.String description, org.hamcrest.Matcher<T> matcher, T... values) {
     return org.hamcrest.core.DescribedAs.describedAs(description, matcher, values);
   }
 
@@ -130,7 +130,7 @@ public class CoreMatchers {
   /**
    * Creates a matcher that always matches, regardless of the examined object.
    */
-  public static org.hamcrest.Matcher<java.lang.Object> anything() {
+  public static <T> org.hamcrest.Matcher<T> anything() {
     return org.hamcrest.core.IsAnything.anything();
   }
 
@@ -141,7 +141,7 @@ public class CoreMatchers {
    * @param description
    *     a meaningful {@link String} used when describing itself
    */
-  public static org.hamcrest.Matcher<java.lang.Object> anything(java.lang.String description) {
+  public static <T> org.hamcrest.Matcher<T> anything(java.lang.String description) {
     return org.hamcrest.core.IsAnything.anything(description);
   }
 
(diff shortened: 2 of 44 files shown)


sf105:
The interface uses matches(Object) because we don't want to over-restrict how a Matcher is used. We don't necessarily want to bind together where the matcher and matchee are declared. This is equivalent to the equals(Object) method on Object, which obviously could be parameterised but isn't. When experimenting with alternatives, we found that there's a usability problem with generics in practice which can lead to lots of tedious type declarations that add little to the effectiveness of a match.

I appreciate your contribution, but I suggest not making assumptions about the authors' understanding.

Thanks

krichter722:
Thanks for you quick feedback.

> I appreciate your contribution, but I suggest not making assumptions about the authors' understanding.

Then you should write _that_ explanation in the Javadoc comment instead of one that leaves the reader with that impression about the understanding of the author. Especially the finding on use cases where this design has advantages should definitely be listed there because it's very helpful at this location (in case you have it somewhere else, please add a link instead).

> The interface uses matches(Object) because we don't want to over-restrict how a Matcher is used. We don't necessarily want to bind together where the matcher and matchee are declared. This is equivalent to the equals(Object) method on Object, which obviously could be parameterised but isn't. When experimenting with alternatives, we found that there's a usability problem with generics in practice which can lead to lots of tedious type declarations that add little to the effectiveness of a match.

The idea of generics is to avoid casting and eventual exceptions - not so much for the sake of effectiveness, but type-safety and static analysis - the code becomes more effective because it's a lot easier to maintain, not in an algorithmic runtime-oriented sense. Like I explained, the changes have no downside and don't add any overhead for users which want to use `Matcher` instead of `Matcher<T>`, but allow user who wish to write type-safe code - which implies the necessary overhead; e.g. the missing feature to call static methods with type parameter is a real bummer in Java since having to write `Matcher.<String>any()` instead of `any()` can be tedious indeed, however it's not necessary.

krichter722:
@tumbarumba How do you feel about this? I'm using only this branch when I'm using Hamcrest and have to rebase it regularly. The type safety is missing for no reason and requires maintenance overhead in almost all projects using Hamcrest because of necessity of test coverage which would be covered by type restrictions otherwise.

krichter722:
For what it's worth, I'm using [hamkrest](https://github.com/npryce/hamkrest) for projects which I migrated to Kotlin already which solves the problem I was trying to solve here.

Issue #207, opened 2018-05-26
Move to Java 8


nhojpatrick:
Hamcrest future

Bump minimum version of Java to Java 8.

Most projects in my view that are using Java 5, 6 or 7, are stable, they are happy with the features currently provided in Hamcrest v1.3, or would be find using version 1.3 if they wanted to use Hamcrest. Not bumping the version I feel holds back hamcrest as after decades of years between Java versions it is now only 6 months before potential new changes are out and hamcrest should be reworked to handle the sudden step change.

Rebel Labs via Zero Turnaround did a conference, and 86.9% are using Java 8 for their main production application, with 6% Java 7 and 2.8% Java 6 or older.

9:52 mark in this Devoxx UK 2018 video
https://www.youtube.com/watch?v=vPqMWPJNSoA&index=90&list=PLRsbF2sD7JVpRvLpv_Cub94zsM1aHm-Op

sf105:
This is obviously an important move. I think that Java 8 is a big enough change that it probably should have a new version and an API that exploits new features. A first step, however, would be to get everything to compile. 

What I'm seeing is that more people than you might expect still have Java 7 code, so we should be careful about leaving them behind.

nhojpatrick:
What about starting multiple parallel branches per java version. Then merge into the correct branch and forward and fix as required.
v2.x Java 7 or newer
v3.x Java 8 or newer
v4.x Java 11 or newer [NOTE 1]

Not sure what else to suggest otherwise those that do more to Java 9/10/11 will experience potentially more and more issues and might not be able to use hamcrest depending what the issues are.

NOTE 1 - no point Java 9 as already EOL, Java 10 will be EOL sept 2018 when Java 11 is out and that is LTS. And it might take a few months to decied and agree what approach to take and start implementing changes required.

thespags:
@sf105 what's the biggest hurdle or blockers for updating to java8? Is it technical like overcoming generics or having the capacity to work on it?

If it's technical, I like to better understand the root of the problem. If it's capacity, I keep looking for places where I can help, but I need to see what are the challenges.

sf105:
challenges. 
- I fear we still need to support Java 7, but maybe that could just be frozen. 
- for similar reasons we probably ought to support Java 8 for a while.
- it's screaming for a more functional solution, but I haven't quite got the confidence about how to get there from here.

I like @nhojpatrick's numbering scheme.

schernov-xo:
[...]

Issue #206, opened 2018-05-26
Java 8 Supporting Language Features


nhojpatrick:
Hamcrest future

Java 8 introduced DateTime, Optional and Streams to make but a few. I believe these should be native matchers within hamcrest.

sf105:
Agreed for a Java 8 version, also closures.

sf105:
I'm not sure how to handle this as there are still plenty of Java 7 users in the world. Time for a new version?

tumbarumba:
Some discussion in #82 that's worth reviewing

tumbarumba:
Merged #421

PR #211, opened 2018-07-27
Remove build on OpenJDK 7 because it's deprecated

krichter722:


sf105:
This one is tricky because, although Java 7 is deprecated, a great many sites are still using it. Similarly for Java 8.

krichter722:
I proposed this in order to facilitate #214 which you vetoed, but it can be realized by enforcing checkstyle version 6.18 which runs on JDK 7 versions, like I proposed in #216. The rulesets format changes with 7.x or 8.x which requires to update a small number of rules as soon as builds on JDK 7 is no longer supported and the checkstyle version upgraded.

krichter722:
I think that it's not overkill to only support JDK version >= 8 for building while they can still be used with versions < 8. Users who still run their applications which use JavaHamcrest using JDK version <8 will not notice the change. Oracle and OpenJDK 8 are long term support versions afaik (like version 11) and will be supported for a long time while 9 is already deprecated again. Up-to-date versions are supported by Travis CI only through Docker.

sf105:
I agree with you for desktop/server. There's still the problem of the Android users (I've had no luck finding someone there to talk to). I'm increasingly convinced that we should split the project into <=7, and >=8 builds. Noone seems to have noticed the new java-hamcrest artefact, so time to revert to upgrading hamcrest-all?

krichter722:
@sf105 FYI, I rebased this PR to remove all old Java versions (11 already is deprecated, but might be in use) and include the latest OpenJDK versions in Travis. Maybe you've changed your mind about this. I didn't stumble over JDK <= 7 for years, now.

sf105:
Again, Joe's call. I'm at a client that is in the process of writing a lot of new code in Java 8 (!). Maybe it is time to drop 7 as no-one in the Android world will even answer an email.

tumbarumba:
I'm inclined to leave the Java 7 dependency in for the moment. I don't have a clear picture of the full scope of where people use Hamcrest, and what their needs are. Having said that, I think that Hamcrest needs a _very_ conservative policy on backwards compatibility. It interesting to note that [JUnit 4.13](https://github.com/junit-team/junit4/issues/1496) _still_ supports Java 5, even though that has been unsupported for years.

Even with the above, I think Hamcrest needs to grow, and adopt more modern language features. 
Issues #206 and #207 are a good place to discuss exactly how we do that (there are already some good suggestions in those issues).

I'm closing this PR, and we can continue discussion in the other issues mentioned.

PR #214, opened 2018-07-28
Add checkstyle setup

krichter722:
Requires removal of OpenJDK (or an older checkstyle version which I don't recommend) proposed in #211.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,9 @@
 language: java
 
-# Need to select an older Ubuntu distribution that supports JDK7
-dist: trusty
-
-# Workaround to using openjdk7 with Gradle due to security issue:
-# https://github.com/gradle/gradle/issues/2421
-before_install:
-- BCPROV_FILENAME=bcprov-ext-jdk15on-158.jar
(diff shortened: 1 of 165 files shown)


sf105:
Thanks for putting all this effort it. I'm afraid I'm not keen on checkstyle for a small library maintained by a few people. Some of the warnings are useful, but some are not. For example, I think it's fine to have very small methods on a single line. If we wanted to use checkstyle, I would start with a skeleton implementation and add rules incrementally. In the meantime, I tend to rely on Idea's checking.

krichter722:
@sf105 I kept this alive since I'm maintaining a patch of Hamcrest which you refused to merge at some point. I'm providing an update in order to share, not push this PR. I can understand that you want to have your own style. Since most of the work is already done, it's probably easier to go from this patch to create your own rules rather than start your own. As long as you don't enforce a certain code style, you can't control technical debt piling up.

tumbarumba:
I've merged #217. I'm not keen on this PR, as I think it brings in too much.

krichter722:
That's fine, you can use it to get inspiration for rules to be added step by step. Static analysis is a one way to quality :)

PR #215, opened 2018-07-28
Change java.io.File to java.nio.file methods

krichter722:
java.nio.file methods avoid loosing errornous states because of boolean
return values which need to be checked whereas IOException thrown by
java.nio.file methods are enforced by the compiler.


sf105:
You make a good point about catching failures. It turns out that some operations are not available on all platforms so I ended up reverting, but adding result checks where appropriate. Thanks

PR #216, opened 2018-08-06
Checker header in checkstyle

krichter722:
As suggested in #214 the checkstyle setup can be put forward step by
step in order to minimize controversy. The start contains a rule to
enforce a license header on every Java files which is a must-have.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,6 +1,6 @@
 BSD License
 
-Copyright (c) 2000-2015 www.hamcrest.org
+Copyright (c) 2000-2022 www.hamcrest.org
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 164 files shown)


sf105:
Sorry to put you through all this. I'm not sure I see the need to enforce that one. It seems to me that many projects have a short header that references the project licence.

krichter722:
@tumbarumba What do you think about the headers? Smaller headers are easier to handle, however most IDE will hide them anyway, so it won't make a difference.

nhojpatrick:
@krichter722 happy to look at getting merged, a huge boilerplate copyright text at the top of a file seams pointless, so personally would prefer shorter snipper just pointing to a LICENSE file in the root of the repo.

will aim get to get something merged by the end of summer...

sf105:
I agree with @nhojpatrick. We really don't need this kind of noise in a little library like Hamcrest

krichter722:
Can't get checkstyle to work with Groovy-Gradle. The config and the headers are on the branch in case someone wants to take over.

PR #217, opened 2018-08-08
Start checkstyle setup 2

krichter722:
As suggested in #214 the checkstyle setup can be put forward step by
step in order to minimize controversy. The start contains a rule to
enforce spaces instead of tab characters which are already used in
roughly 99.99% of the source.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,6 +6,7 @@ group = "org.hamcrest"
 version = "2.2-SNAPSHOT"
 
 subprojects {
+    apply plugin: 'checkstyle'
     apply plugin: 'java-library'
 
     sourceCompatibility = JavaVersion.VERSION_1_7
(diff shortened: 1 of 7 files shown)


krichter722:
@sf105 This could be a start for checkstyle. You could collect experience over time.

tumbarumba:
As a simple start, I like this, and it shows the value of having a check like this (some tab characters have snuck into the codebase).

Are there any objections to merging this?

sf105:
As Karl-Philipp and I discussed already, I'm not keen to add checkstyle to the build as, at the moment it's very simple with hardly any dependencies. I'm not convinced that it brings enough benefit for a lot of imports (it's Joe Armstrong's old line about wanting the banana but getting the whole jungle). That said, this is your show now, so you make the call.

tumbarumba:
Calling! It's in! This happens to be a jungle I happen to like. As the random tabs show, keeping consistency in a codebase with a number of different contributors can be tricky. I think tools like this help with consistency.

PR #222, opened 2018-10-28
Merge gh-pages to master

tumbarumba:
This moves the root of the gh-pages branch to the `docs` folder on master. Once merged, the [JavaHamcrest/settings](https://github.com/hamcrest/JavaHamcrest/settings) on GitHub can be changed so that the GitHub Pages source is `master branch \docs folder`. After changing this, you can test that the website at http://hamcrest.org/JavaHamcrest still works. The `gh-pages` branch should be deleted at this stage to remove any confusion from casual onlookers about where the documentation is stored.


sf105:
Config changed.

PR #223, opened 2018-11-01
Update links to Tutorial and Distributables

tumbarumba:
I've replace links to two outdated pages on the old Google Code site. The replacement pages are currently on the JavaHamcrest wiki on GitHub.

For the tutorial, I've linked to the existing Tutorial on the wiki. It's still out of date, but I plan to refresh it soon.

For the distributables, I've created a new wiki page. I'd appreciate it if people could review the page for accuracy and improvements.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -6,7 +6,7 @@
   <meta name="description" content="Hamcrest"/>
   <meta name="keywords" content="hamcrest java matchers junit testing"/>
   <meta name="robots" content="index, follow"/>
-  <meta name="date" content="2012-09-01"/>
+  <meta name="date" content="2018-11-01"/>
   <title>Java Hamcrest</title>
 
   <style>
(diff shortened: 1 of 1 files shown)


sf105:
Just realised the target is the wiki. Should we be using /doc pages so they're versioned?

tumbarumba:
We could do. I was thinking about doing that, but then I noticed that there was already a version of the tutorial in the wiki, so I linked to that instead.

I'm happy to move the tutorial and distribution pages into the github pages, but it will take me a while to get around to that (probably not until next week). I'm actually planning to upgrade the tutorial to use JUnit 5 (and possible another example in JUnit 4).

How about you merge this as an incremental improvement, and I can create a new pull request later that pulls the wiki pages into the `/docs` folder?

sf105:
agreed

PR #226, opened 2018-11-05
Add FloatIsCloseTo matcher

laz2:
* Add `FloatIsCloseTo` matcher as copy of `IsCloseTo` matcher
* Rename `IsCloseTo` matcher to `DoubleIsCloseTo`
* Change error message for excluding errors in additional floating calculations (May user catch this error even with doubles?)
* Ignore eclipse specific files

PS.
Example of error with additional floating calculations:

```java
public static void main(String[] args) {
        System.out.println(abs(1.0f - 0.1f) - 0.5f);
        System.out.println(abs(1.0d - 0.1d) - 0.5d);
}
```
After run:
```
0.39999998
0.4
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@
 gradle.properties
 release-key.asc
 grade-wrapper.properties
+/.settings
+/.project
+/.classpath
--- a/hamcrest-library/src/main/java/org/hamcrest/Matchers.java
+++ b/hamcrest-library/src/main/java/org/hamcrest/Matchers.java
@@ -1262,7 +1262,22 @@ public class Matchers {
    *     the delta (+/-) within which matches will be allowed
    */
   public static org.hamcrest.Matcher<java.lang.Double> closeTo(double operand, double error) {
-    return org.hamcrest.number.IsCloseTo.closeTo(operand, error);
+    return org.hamcrest.number.DoubleIsCloseTo.closeTo(operand, error);
+  }
+
+  /**
+   * Creates a matcher of {@link Float}s that matches when an examined float is equal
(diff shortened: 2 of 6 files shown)


laz2:
> First, what problem are you trying to solve, what is the motivation for this change?

There's no ``closeTo`` matcher for float value (only with casting to double).

> Secondly, this would be a breaking change for existing users of IsCloseTo.

I can rename ``DoubleIsCloseTo`` -> ``IsCloseTo``. I tried resolve issue with single ``IsCloseTo`` and generics, but that solution doesn't work.

sf105:
Sorry, to be clearer, is this just a matter of inconvenience, or is there a real problem? Are you seeing significant differences in the results when casting floats to doubles?

laz2:
> Are you seeing significant differences in the results when casting floats to doubles?

No, I don't. I think there may be problem. And of course there'r ugly type cast in strong type system.

nhojpatrick:
@laz2 i personally prefer using the Matchers directly like `Is.is` instead of just referring to `CoreMatchers.is`. like microservices instead of monoliths, instead of one IsCloseTo to rule them all, creating FloatIsCloseTo and DoubleIsCloseTo makes more sense, also you might want to enforce Float or Double and not want auto boxing or type conversion.

sf105:
A long time ago, I had a plan to extract related matchers into a set of common DSL types, which I think might work here. Double and Float is probably one of the few occasions where a static import doesn't work.

nhojpatrick:

laz2:
Done in another pull request #385

Issue #224, opened 2018-11-05
Release Version 2.1 On Maven


ankurpathak:
I can see the last artificat of Hamcrest released on maven central was back  in 2015.
But the latest master has many updates which are not availabe on that maven build.
For example empty matcher for Map. So please release the binaries for latest master
on maven central. So we can consume it in our projects. Also the latest master and
binaries on maven central should be in sync.

tumbarumba:
I think this is a great idea. It's been far too long since releases. I plan to try and do this myself over the next couple of weeks.

@hamcrest/hamcrest-java-devs: looking at the shiny new [Hamcrest Distributables](http://hamcrest.org/JavaHamcrest/distributables) page, there are 5 jars mentioned. I can see the code for `hamcrest-core.jar` and `hamcrest-library.jar`, but I don't see any code for `hamcrest-integration.jar` or `hamcrest-generator.jar` (I understand `hamcrest-all.jar` is just the combination of all the other jars). Does anyone have any idea where this missing code might be? Is it still relevant?

sf105:
I tried to collapse everything into one jar. We don't need the generator any more as we've stopped doing the code generation for the Matchers class. There's still a decision about whether to have separate jars for the test framework integrations as they're outside the JDK. Might make life easier for dealing with junit 4/5 and other test frameworks.

tumbarumba:
Thanks @sf105. The test frame integration classes do not appear to be checked into this github repository. Do you have them somewhere else?

tumbarumba:
I found the java files on the old Google Code repo: https://code.google.com/archive/p/hamcrest/source/default/source

I'll see if I can import them into GitHub this weekend.

tumbarumba:
@hamcrest/hamcrest-java-devs: I've just landed #228, which is preparing the Gradle build for reproducing the various jars as per the 1.3 release. Included in this change: I've set the version to `1.4-SNAPSHOT`.

Are there any objections to calling the next version `1.4`? If so, what _should_ the next version be?

ankurpathak:
@tumbarumb  Using 1.4 for new version will create same confusion which to use 1.3.x.x or 2.x.x.x. So lets
skip these 1 and 2. And start a new life for hamcrest with 3 or 4 and regular releases on maven. So their 
wont be any confusion remains. Since human by their greedy nature have the tendncy to use latest and greatest one.

tumbarumba:
[...]

PR #228, opened 2018-11-10
 Build separate jars for core, library, and all.

tumbarumba:
This is done by converting to a multi-project gradle build, with subprojects for `hamcrest-core` and `hamcrest-library`. The root project is responsible for combining all the individual jars into `hamcrest-all.jar`, as well as configuring the publishing.

Note that this changes the version from '2.0.0.0' to '1.4-SNAPSHOT'

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,11 @@
-/build
-/classes
-/bin
-/out
+build/
+classes/
+bin/
+out/
 .DS_Store
 *~
 *.swp
-/.idea
+.idea/
 *.iws
 *.ipr
 *.iml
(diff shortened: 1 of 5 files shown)


sf105:
I thought we were just going to have one big jar and drop the smaller ones? Basically, everything that just depends on the JVM ad there’s not much point in distinguishing between core and library.

tumbarumba:
Ah, I misunderstood the conversation last Tuesday, where we were discussing keeping APIs the same as 1.3 (it was in a pub and quite loud!), as well as your comments on Issue #224, where you were keeping hamcrest-integration as a separate jar (it has dependencies outside the jdk). For some reason I had told myself you meant that *all* the jar packages should stay the same.

I can change this, though I think we still need the multi-project aspect. If I bring the integration code back from Google Code, we can keep that as a separate jar, and merge core and library into a single jar, as you had before this specific change.

How does that sound?

PR #229, opened 2018-11-13
Merge core and library, revive hamcrest-integration

tumbarumba:
This is a big change to the project structure, but there are **no** changes to any Java code.

After chatting with @sf105 on Issue #224, I reworked the way that the code is organised and packaged. The more I think about it, the less I like the idea of preserving the same packaging. It's been such a long time between releases, I don't think there's a huge benefit in keeping the old packaging around.

As this is a breaking change due to the repackaging, the version has changed from `1.4-SNAPSHOT` to `2.1-SNAPSHOT`. I realise that the Java API hasn't actually changed, but since there is already a version `2.0.0.0` published on Maven Central, the next version published must be incrementally higher than that, at a minimum.

I have revived hamcrest-integration as separate subproject. This was done by unzipping `hamcrest-integration-1.3.jar` downloaded from Maven Central. Unfortunately, the corresponding code available from the archived hamcrest-java on Google Code appeared out of date. I could not see test code for java-integration (I had a poke around the Google Code archive, but no luck there).
    
Salient points:
* New project `hamcrest` is the result of merging `hamcrest-core` and `hamcrest-library`.
* `hamcrest-integration` remains an independent project.
* `hamcrest-integration` remains only as a 1.3 line bugfix backstop. It is not published, and its version is pinned at `1.3.1-SNAPSHOT`.
* All aggregated jars removed (no `hamcrest-all` or `java-hamcrest`), which means only a single artefact is published: `org.hamcrest:hamcrest:<version>` (jar, sources and javadoc)


sf105:
I would have thought this would be version 1.4, rather than 2? It's a change in the packaging, rather than breaking code.

tumbarumba:
That's true, @sf105. However, we have already released a version 2.0.0.0 of *something*. Unfortunately, we are unable to remove it from Maven Central. I think it would be highly confusing to have a version 1.4 artifact that is more recent than a version 2.0.0.0 artifact, even it it was a mistake.

PR #244, opened 2019-01-04
Allow "normal" objects to describe themself, e.g. in case of a mismatch

PirolA:
*Motivation*
I needed to give detailed descriptions of an object in case of a mismatch and couldn't use toString(). 

*Solution*
* Use the existing `SelfDescribing` Interface and retrieve a description based on its `describeTo` method.

### DIFF
--- a/hamcrest/src/main/java/org/hamcrest/BaseDescription.java
+++ b/hamcrest/src/main/java/org/hamcrest/BaseDescription.java
@@ -3,6 +3,8 @@ package org.hamcrest;
 import org.hamcrest.internal.ArrayIterator;
 import org.hamcrest.internal.SelfDescribingValueIterator;
 
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.util.Arrays;
 import java.util.Iterator;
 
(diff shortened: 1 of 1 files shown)


sf105:
What is the situation where you can't use the `appendDescriptionOf` method?

tumbarumba:
This change seems like an improvement. @PirolA, can you add a test for the changed functionality to this PR?

hamcrest/src/main/java/org/hamcrest/BaseDescription.java
```diff
@@ -29,6 +31,19 @@ public Description appendDescriptionOf(SelfDescribing value) {
     public Description appendValue(Object value) {
         if (value == null) {
             append("null");
+        } else if (value instanceof SelfDescribing){
```
panchenko: There is a fundamental problem with this approach: normally hamcrest would be a test dependency.

nhojpatrick:
@PirolA interesting idea, as some previous comments have highlights it would mean hamcrest being a non compile dependency. maybe looking at how `spotbugsfindbugs` have `spotbugs-annotations/findbugs-annotations`, where we could support something like thats a making a lightweight runtime jar if you really needed.

nhojpatrick:

PR #245, opened 2019-01-10
AllOf/AnyOf: Pass the matchers to constructor using varargs

dgroup:
Add a way to construct the `AllOf`, `AnyOf` matchers using the `varargs`.
Some projects prohibit the using of static imports/methods, therefore we have to wrap the matchers to the `Iterable`, it's not convenient and looks a bit ugly.

@tumbarumba , could you please review/give feedback on this pull request?


tumbarumba:
Nice one

Issue #249, opened 2019-03-03
IsUnmodifiable matcher feature


Andremoniy:
It would be nice to have as a standard feature a special matcher to verify that a given collection is unmodifiable.

This sort of checks raises when we test certain type of methods which should return unmodifiable implementation rather than regular one.

PR #250, opened 2019-03-03
IsUnmodifiableCollection matcher feature #249

Andremoniy:
My proposal of implementation of the issue #249 .

Should work fine for both `Collections::unmodifiable*` implementations, and guava `Immutable*` collections.

### DIFF
--- a/hamcrest/src/main/java/org/hamcrest/collection/IsUnmodifiableCollection.java
+++ b/hamcrest/src/main/java/org/hamcrest/collection/IsUnmodifiableCollection.java
@@ -1,0 +1,176 @@
+package org.hamcrest.collection;
+
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeDiagnosingMatcher;
+
+import java.lang.reflect.Array;
+import java.lang.reflect.Constructor;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Matches if collection is truly unmodifiable
+ */
+public class IsUnmodifiableCollection<E> extends TypeSafeDiagnosingMatcher<Collection<? extends E>> {
+
+    private static final Map<Class, Object> DEFAULT_COLLECTIONS = new HashMap<>();
+
+    static {
+        final List<String> list = Arrays.asList("a", "b", "c");
+        DEFAULT_COLLECTIONS.put(Collection.class, list);
+        DEFAULT_COLLECTIONS.put(List.class, list);
+        DEFAULT_COLLECTIONS.put(Set.class, new HashSet<>(list));
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    protected boolean matchesSafely(final Collection collection, final Description mismatchDescription) {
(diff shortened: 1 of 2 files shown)


Andremoniy:
I'm not sure why the build fails for OpenJDK 7. It gives the following error:

> Gradle 5.2 requires Java 8 or later to run. You are currently using Java 7.

However I don't think I am using anything specific for Java version >= 1.8, neither specifying Gradle version explicitly.

tumbarumba:
@Andremoniy Gradle error was my fault - I committed a Gradle upgrade to master without going through PR because I thought it was safe (hangs head in shame). I've just reverted that commit on master.

tumbarumba:
Hi @Andremoniy,

Thanks for spending time on this!

I've had a quick look over your changes. I have a  concern: this code verifies an collection is immutable by trying to mutate it. I think this is a dangerous thing to do, as we don't necessarily know if that will break other things. For example, modifying a shared test fixture might break other tests.

I think the [Mutability Detector](https://github.com/MutabilityDetector/MutabilityDetector) uses a safer approach through reflection (though much more complex).

Have you tried Mutability Detector? Does it solve your needs?

Andremoniy:
> this code verifies an collection is immutable by trying to mutate it

Hi @tumbarumba!
Thank you so much for coming back to my PR.
I found this your remark:

> this code verifies an collection is immutable by trying to mutate it

as very valuable and right. I completely agree, that this is not the right way of doing this check.
I will check what you have proposed and come back shortly with a new solution.

Regards,
Andrey.

Andremoniy:
Hi @tumbarumba !
I've come "shortly" after half of a year :-) 
To avoid brining any new dependencies into the project, I've tried to solve the issue by instantiating a new instance of the same class as a testable collection and perform all checks on this new instance. 
I am pretty sure the tests cases in `IsUnmodifiableCollectionTest` can be augmented; if you have any ideas, I would love to add them as well.

nhojpatrick:
@Andremoniy interesting idea, probably a good idea to include so you can check and api doesn't expose internal or writable collection

Andremoniy:
@nhojpatrick I've rebased my branch, though I am not sure if I did in the most righteous way, because it now shows me all merged commits as part of my PR. Please let me know if I need to re-do something.

nhojpatrick:

Andremoniy:
> Going to try and kick start hamcrest, so if you want to get it merged, please rebase from the branch `v2.3-candidates`. Still trying to understand how has permissions to perform a release.

@nhojpatrick  I've rebased it from the branch `v2.3-candidates` as requested and also changed the base of the PR (I guess that's what I should have done as well?).

nhojpatrick:
@Andremoniy
`v2.3-candidates` was just a stable branch/tag, and I don't think you can change a PR target once it's been open which isn't great.
Looking at it now and will try and sort ou.

Andremoniy:
> @Andremoniy `v2.3-candidates` was just a stable branch/tag, and I don't think you can change a PR target once it's been open which isn't great. Looking at it now and will try and sort ou.

I feel like I did something wrong and now I am confused.

nhojpatrick:
@Andremoniy thanks merged, hopefully release won't be that long away...

nhojpatrick:
i knew i'ld potentially be review and merging, so wanted to give a static instead of master, which itself might have constantly updated, so out of date by the time i get it, and then require another rebase...

Andremoniy:
@nhojpatrick is there any chance to reach you out by Skype/hangouts?

Issue #248, opened 2019-02-25
containsInAnyOrder can depend on order


Gnurfos:
The algorithm goes like: for each item, try each matcher in the given order and consume the first one that matches.

This causes the following test to fail:

```
    @Test
    public void containsInAnyOrderFailsToTryAllCombinations() {
        assertThat(
                Arrays.asList("foo", "friend"),
                containsInAnyOrder(
                        containsString("f"),
                        containsString("o")
                )
        );
    }
```

The doc says:

> N.B. each of the specified matchers will only be used once during a given examination, so be careful when specifying matchers that may be satisfied by more than one entry in an examined iterable

So that could explain it, but it's still surprising for the user.

brownian-motion:
This seems like greedy consumption of elements. It could be addressed by backtracking, but I'm not sure what the performance of that would be like, and it would certainly break that contract

sf105:
This is a good point. I expect someone out there has a clever algorithmic solution...

Gnurfos:
I don't know if it is clever, but there is a pull request with some (rather naive) algorithm.

PR #251, opened 2019-03-05
Make IsIterableContainingInAnyOrder try all matcher permutations

Gnurfos:
This is a fix for issue #248
I'm not sure about the names I used, but could not find better ones. Don't hesitate to propose some.

This introduces some loss of information in mismatch reports in some cases (which precise matchers or items remained unpaired), but on the other hand, the full list of items is printed all the time (which, from my experience, is more useful to the user, in non-ordered cases).

This information could be reintroduced if needed, at the cost of some added complexity and performance.

### DIFF
--- a/hamcrest/src/main/java/org/hamcrest/collection/IsIterableContainingInAnyOrder.java
+++ b/hamcrest/src/main/java/org/hamcrest/collection/IsIterableContainingInAnyOrder.java
@@ -3,11 +3,9 @@ package org.hamcrest.collection;
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
 import org.hamcrest.TypeSafeDiagnosingMatcher;
+import org.hamcrest.internal.InPlacePermutator;
 
-import java.util.ArrayList;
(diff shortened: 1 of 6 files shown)


tumbarumba:
This looks like a lot of extra complexity for not a lot of benefit.

What problem does this fix? Is there a specific use case this change addresses?

tumbarumba:
Ah! I just read #248, I see the use case you're concerned about

nhojpatrick:
@Gnurfos think i understand the issue so should probably be merged.

nhojpatrick:

Gnurfos:
@FichteFoll I agree with all the concerns you listed above. What I did at the time was fitting my needs, and I valued correctness of the test above all else, but I understand that it's not good enough (reporting and performance) for the general case.

Unfortunately, I'm not into Java at the moment, and won't be able to put efforts into this in any foreseeable time.
Feel free to close this pull request, as I'm afraid not much of it would be of use, with a different algorithm.

PR #255, opened 2019-04-17
Fix #254 Improve description when byte value is mismatched

### LINKED ISSUE #254: AssertionError message is unhelpful when match fails for byte type
Here is an example of a JUnit 5 test displaying this issue:
```
import static org.hamcrest.Matchers.equalTo;
import static org.hamcrest.Matchers.is;

import org.junit.jupiter.api.Test;

public class ExampleTest {

	@Test
	void testByte() {
		assertThat((byte) 5, is(equalTo(5))); // Error message is bad
	}

	@Test
	void testShort() {
		assertThat((short) 5, is(equalTo(5))); // Error message is good
	}
}
```
From `testShort` the error message is good: (Note: <5**s**>)
```
Expected: is <5>
     but: was <5s>
```
But from `testByte` you get:
```
Expected: is <5>
     but: was <5>
```
Which looks ok `<5>` is `<5>`! Of course the reason the failed is the types are mismatched so the failure is correct, but it would be good to show this more clearly in the message.

jamesmudd:


tumbarumba:
Looks good, thanks @jamesmudd

PR #257, opened 2019-04-24
Use "<? super T>" for all arguments of anyOf

### LINKED ISSUE #256: Matchers.anyOf: inference variable has incompatible bounds with Java >8
The following code sample works with OpenJDK 8 but does not compile with OpenJDK 11:
```
public class SimpleTest {
	@Test
	public void test() {
		List<String> list = new ArrayList<>();
		assertThat(list, anyOf(hasItem("a"), hasItem("b")));
	}
}
```

Compiling with Java 11 fails with
```
    [javac]     method Assert.<T#1>assertThat(String,T#1,Matcher<? super T#1>) is not applicable
    [javac]       (cannot infer type-variable(s) T#1
    [javac]         (actual and formal argument lists differ in length))
    [javac]     method Assert.<T#2>assertThat(T#2,Matcher<? super T#2>) is not applicable
    [javac]       (inference variable T#3 has incompatible bounds
    [javac]         lower bounds: String,Object
    [javac]         lower bounds: String,Object)
    [javac]   where T#1,T#2,T#3 are type-variables:
    [javac]     T#1 extends Object declared in method <T#1>assertThat(String,T#1,Matcher<? super T#1>)
    [javac]     T#2 extends Object declared in method <T#2>assertThat(T#2,Matcher<? super T#2>)
    [javac]     T#3 extends Object declared in method <T#3>hasItem(T#3)
```

If the first argument of the Matchers.anyOf methods would be changed from Matcher<T> to Matcher<? super T> (like in Matchers.allOf), it would work.

phoenix384:
fixes #256

### DIFF
--- a/hamcrest/src/main/java/org/hamcrest/Matchers.java
+++ b/hamcrest/src/main/java/org/hamcrest/Matchers.java
@@ -98,7 +98,7 @@ public class Matchers {
    * For example:
    * <pre>assertThat("myValue", anyOf(startsWith("foo"), containsString("Val")))</pre>
    */
-  public static <T> org.hamcrest.core.AnyOf<T> anyOf(org.hamcrest.Matcher<T> first, org.hamcrest.Matcher<? super T> second) {
+  public static <T> org.hamcrest.core.AnyOf<T> anyOf(org.hamcrest.Matcher<? super T> first, org.hamcrest.Matcher<? super T> second) {
     return org.hamcrest.core.AnyOf.anyOf(first, second);
   }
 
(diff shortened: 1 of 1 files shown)


tumbarumba:
Thanks @phoenix384, I was able to reproduce your issue after a bit of effort. Because Hamcrest itself is compiled with `sourceCompatibility = JavaVersion.VERSION_1_7`, the problem doesn't appear in that project. However, if I set up a new project with a Hamcrest as a dependency, and use `sourceCompatibility = JavaVersion.VERSION_1_8`, JDK 1.11 couldn't compile.

PR #258, opened 2019-05-21
Add relativelyCloseTo implementation

sooniln:


nhojpatrick:
@sooniln source branch appears to have been deleted so won't be able to merge.

if you want to rebase from master, as `hamcrest-core` and `hamcrest-library` have been refactored a lot and also deprecated, so that everything is just in `hamcrest`. we can try again

PR #261, opened 2019-06-05
Introduce HasGetterWithValue

retienne:


retienne:
Did not know that it must be Java7 compatile

PR #267, opened 2019-07-24
Use platform-specific line separators

albertdev:
I noticed that when I log an assertion error created by Hamcrest that it uses a single line feed character (Unix style end-of-line). Since the rest of my log uses carriage return / line feed pairs (Windows style end-of-line) this will confuse some editors like Vim or Eclipse which generally expect a file to use the same style throughout.

Rather than hardcoding one or the other, the `System.lineSeparator()` function will return the end of line separator appropriate for the current runtime, i.e. "\r\n" on Windows and "\n" on Unix. This way it will adapt to the Java runtime's platform. And should for some reason a user prefer the other style, they should be able to set the `line.separator` property at startup.

### DIFF
--- a/hamcrest/src/main/java/org/hamcrest/MatcherAssert.java
+++ b/hamcrest/src/main/java/org/hamcrest/MatcherAssert.java
@@ -10,9 +10,11 @@ public class MatcherAssert {
         if (!matcher.matches(actual)) {
             Description description = new StringDescription();
             description.appendText(reason)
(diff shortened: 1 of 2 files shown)


tumbarumba:
I've just fixed a problem with TravisCI that stopped the checks running. Can you please rebase this PR?

albertdev:
@tumbarumba Done.

PR #275, opened 2019-10-11
Changed signature of method containsInAnyOrder to allow avoid calling…

yatsenko-ihor:
Changed signature of method containsInAnyOrder to allow avoid calling toArray() method or casting to the array and build next construction

List<String> actual = Arrays.asList("a", "b", "c");
List<String> expected = Arrays.asList("a", "c");
assertThat(list, containsInAnyOrder(expected))

Some results from the internet with recommendations to call toArray():
https://stackoverflow.com/questions/21624592/hamcrest-compare-collections
http://haroonzone.github.io/comparing-lists-with-hamcrest/ (ListMatcher2.java)

### DIFF
--- a/hamcrest/src/main/java/org/hamcrest/Matchers.java
+++ b/hamcrest/src/main/java/org/hamcrest/Matchers.java
@@ -1000,11 +1000,11 @@ public class Matchers {
    * <p>For example:</p>
    * <pre>assertThat(Arrays.asList("foo", "bar"), containsInAnyOrder(Arrays.asList(equalTo("bar"), equalTo("foo"))))</pre>
    * 
-   * @param itemMatchers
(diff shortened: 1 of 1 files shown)


tumbarumba:
I really like they way that this simplifies the calling code for collections of items.

Without knowing if this is true, I also worry that this will break other code that is passing collections of matchers. Can you add another test to `IsIterableContainingInAnyOrderTest` to verify that this doesn't cause a regression?

nhojpatrick:
@yatsenko-ihor looking at closing pr and getting merged, looks a good idea

can you keep the current signature but mark as deprecated as i'm sure if we do this change we might get bugs as a result if people are using existing signature

nhojpatrick:

PR #282, opened 2020-01-12
Changed signature of Matcher.matches(Object) to matches(T) using the parameter of the type.

### LINKED ISSUE #199: Matcher.matches needs to have parameterized argument

krichter722:
Unused tests which are no longer necessary have been removed.

Closes #199

This has already been discussed in #200 without a satisfying outcome: I'm using only this branch when I use Hamcrest and have to rebase it regularly. The type safety is missing for no reason and requires maintenance overhead in almost all projects using Hamcrest because of necessity of test coverage which would be covered by type restrictions otherwise.

### DIFF
--- a/hamcrest/src/main/java/org/hamcrest/BaseMatcher.java
+++ b/hamcrest/src/main/java/org/hamcrest/BaseMatcher.java
@@ -17,7 +17,7 @@ public abstract class BaseMatcher<T> implements Matcher<T> {
     }
 
     @Override
-    public void describeMismatch(Object item, Description description) {
+    public void describeMismatch(T item, Description description) {
         description.appendText("was ").appendValue(item);
     }
 
--- a/hamcrest/src/main/java/org/hamcrest/CoreMatchers.java
+++ b/hamcrest/src/main/java/org/hamcrest/CoreMatchers.java
@@ -75,7 +75,7 @@ public class CoreMatchers {
    * @param values
    *     optional values to insert into the tokenised description
    */
-  public static <T> org.hamcrest.Matcher<T> describedAs(java.lang.String description, org.hamcrest.Matcher<T> matcher, java.lang.Object... values) {
+  public static <T> org.hamcrest.Matcher<T> describedAs(java.lang.String description, org.hamcrest.Matcher<T> matcher, T... values) {
     return org.hamcrest.core.DescribedAs.describedAs(description, matcher, values);
   }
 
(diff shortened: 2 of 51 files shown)


nhojpatrick:
@krichter722 looks good, wondering about if we need to keep the older signatures in the short term

krichter722:
Someone needs to figure out the commented out tests first. The syntax issues remain. There's Kotlin to the rescue.

Issue #152, opened 2016-06-24
hasProperty with value fails if the scope of the instance's class is not public


JasonWThompson:
If an instance of a default scoped class is passed into the hasProperty matcher, the assertion will always fail when the get method is invoked to get the value of the property.

trejkaz:
I believe this is the issue I'm seeing? My test case:

```
import org.junit.Test;

import java.nio.file.Path;
import java.nio.file.Paths;

import static org.hamcrest.Matchers.*;
import static org.junit.Assert.assertThat;
import static org.junit.Assert.assertTrue;

public class LibraryTest {
    @Test
    public void testSomeLibraryMethod() {
        Path path = Paths.get("hi.txt");
        assertThat(path.getFileName().toString(), is(equalTo("hi.txt")));
        assertThat(path, hasProperty("fileName", hasToString("hi.txt")));
    }
}
```

PR #286, opened 2020-03-13
Failing test for #152

trejkaz:
I've added a test that I think should pass, but it fails. Now I just have to figure out what is being done wrong.


nhojpatrick:
@trejkaz looks good, please can you rebase from master, as `hamcrest-core` and `hamcrest-library` have been refactored a lot and also deprecated, so that everything is just in `hamcrest`.

trejkaz:
Rebase seemed suspiciously clean but I guess I'll let Travis be the judge.

trejkaz:
It failed, but then when I re-read my ticket description above that is the expected, yes. So the question is now how to get the new test to pass. :)

hakanai:
Probably not a lot of benefit getting this into a release because it still wants the actual fix.

PR #291, opened 2020-04-12
Added visibility matchers to check that a class isPublic(), etc.

brownian-motion:
Covers all 4 visibilities, and all reflective elements (classes, methods, fields, etc.) with reasonable messages when used incorrectly.

These matchers are helpful, for example, when enforcing the scope of a public-facing API with a test,
which provides stronger documentation for the future than mere comments.

I've performed similar checks in the past in projects with many, many artifacts to make sure that certain methods are definitely visible outside of a library, for consumption by another library or for invocation by reflective techniques.


brownian-motion:
Re-wrote the implementation to be much, much smaller.

Code coverage is now 100%, and the public API is organized in one class rather than in 4 different ones.

brownian-motion:
@scarytom , @sf105 , or @tumbarumba , would you please consider reviewing this PR when convenient? I understand that maintaining this and other repositories requires a significant amount of your attention, on top of your work outside of this project, and I would very much appreciate your time if any of you are willing to offer it.

sf105:
Interesting idea. Would this make more sense in a package called reflection?

sf105:
Do we need the extra VisibilityUtils class? Utils is often a smell

brownian-motion:
@sf105 Thank you for taking a look. I agree, it would make sense in a `reflection` package. I agree also that the Utils class could be removed, probably by merging it directly into the Enum or *Matchers class. What remains now is an artifact of my original design, which used WAY more classes than needed. I'll push an improvement.

nhojpatrick:
@brownian-motion looks good, please can you rebase from master, as `hamcrest-core` and `hamcrest-library` have been refactored a lot and also deprecated, so that everything is just in `hamcrest`.

brownian-motion:
@nhojpatrick I will rebase like you've described. Thanks!

brownian-motion:
Welp. Got a little caught up during the pandemic. I'll clean this up.

brownian-motion:
@nhojpatrick assuming this passes the build checks, would you please take a look at this when convenient? I've rebased, and applied the style settings from the repository checkstyle file.

nhojpatrick:

Issue #172, opened 2017-02-09
hasKey performance issue


Flame239:
Why `IsMapContaining.hasKey(K key)`  is iterating over whole map, instead of calling `Map.containsKey`? Latter is much faster!

Nibsi:
Same problem with `IsIterableContaining.hasItem(T item)`. Instead of wrapping the element in an `equalTo()` matcher, it could check if the iterable is a `Collection` and use the `Collection.contains()` method, or use a separate `IsCollectionContaining` matcher that only accepts `Collection` types and then also performs the test using the `Collection.contains()` method.

tumbarumba:
I just had a look at implementation of [`IsMapContaining.hasKey(K key)`](https://github.com/hamcrest/JavaHamcrest/blob/master/hamcrest/src/main/java/org/hamcrest/collection/IsMapContaining.java#L100). I get the impression that the current implementation is the way it is because of simplicity and consistency. That method has the same structure as every other method in the `IsMapContaining` class. This means there is only a single implementation of the matchesSafely method.

It would certainly be possible to create an optimised matcher especially for the case of matching an expected literal key inside a map, or a literal item inside a collection. It would be a slight amount of extra complexity for the better performance.

Out of interest, is the performance of this matcher causing any problems for your code?

Nibsi:
Not as of yet, I have to admit.

However, I have written a library that relies heavily on Hamcrest that validates method preconditions and I can foresee that some methods accepting maps or collections may be inside heavily used loops.

Normally, such an algorithm would operate in linear time when using hash maps or hash sets, but with the current implementation of these matchers such an algorithm would perform in `O(n * m)` time, where `n` is the number of collections to process and `m` is the number of elements in these collections.

tumbarumba:
Understood. Are you able to submit a PR?

Nibsi:
Sure. I can probably have it up Friday or Monday evening.

vlsi:
@Nibsi , it looks like I'm bitten with the same issue. Have you had a chance to optimize `org.hamcrest.collection.IsMapContaining#matchesSafely` ?

Nibsi:
@tumbarumba @vlsi sorry guys, I completely got sidetracked. Thanks for the reminder. I'm busy over the weekend, but I'll definitely hop to it the coming week. Stay tuned.

brownian-motion:
[...]

PR #292, opened 2020-04-13
Issue 172: optimize map hasKey() and hasEntry()

brownian-motion:
Implements the enhancement in hamcrest/JavaHamcrest#172

Instead of a linear search over all entries, this change leverages Map.containsKey(K key) to check that a map hasKey() or hasEntry().


hamcrest/src/main/java/org/hamcrest/collection/IsMapContaining.java
```diff
@@ -72,9 +72,32 @@ public void describeTo(Description description) {
...
+        public IsMapContainingEntry(K key, V value)
+        {
+            super(equalTo(key), equalTo(value));
+            this.key = key;
+        }
+
+        @Override
+        public boolean matchesSafely(Map<? extends K, ? extends V> map)
+        {
+            return map.containsKey(key) && super.valueMatcher.matches(map.get(key));
```
nibsi, line 97: There are maps that don't allow null keys and (stupidly) throw an exception when you try to check for its existence. You need to account for them here.
```suggestion
            try {
              return map.containsKey(key) && super.valueMatcher.matches(map.get(key));
            catch (NullPointerException ex) {
              return false;
            }
```

brownian-motion:
Thank you for your feedback, @nibsi . I'm afraid I created the branch from my local `master` branch, which included a different change, instead of `origin/master`; I'm embarrassed I missed that before pushing. I'll rebase and push a PR without the unrelated change.

brownian-motion:
I've added some tests that cover the relevant edge cases covered by the code addition. I believe this commit is ready for final review + merging if approved.

sf105:
Interesting question, the null thing is not something I thought of. The question is what is the right behaviour when using null with an inappropriate collection. Personally, I would have expected it to fail, rather than mismatch, as that's what would happen in the production code. I fear that we'll be hiding failure modes. 

Also, is that a copy'n'paste in the comment?

brownian-motion:
@sf105 There's one case I can think of where the safe API is better: asserting that a null key is absent:

`assertThat(new Hashtable<Object, Object>(), not(hasKey(null)));`

I would expect this to pass and continue to the next assertion; allowing a null pointer exception would force you to avoid touching particular types/code paths or start to put conditional logic and try/catches in your tests.



Also, sorry, where do you mean? I'm not sure what copy-and-paste you mean.

brownian-motion:
@sf105 I would be happy to make whichever further changes you think would benefit the project the best. When convenient, please let me know what further adjustments might be appropriate here.

nhojpatrick:
@brownian-motion looks good, please can you rebase from master, as `hamcrest-core` and `hamcrest-library` have been refactored a lot and also deprecated, so that everything is just in `hamcrest`.

brownian-motion:
Hello again! I know it's been two years, but I've rebased this PR and would be happy to help with any further steps to get it ready for merging.

@nhojpatrick please feel free to look through this again when convenient :)

nhojpatrick:

Issue #174, opened 2017-04-22
hasSize and similar matchers to provide more debugging info


alb-i986:
Consider the following test case:
    
    @Test
    public  void asd() {
        assertThat(Arrays.asList(57, 965), hasSize(greaterThan(3)));
    }

When this guy fails, we'll get:

    java.lang.AssertionError: 
      Expected: a collection with size a value greater than <3>
        but: collection size <2> was less than <3>

Which is ok, but it doesn't provide much info in order to debug the failure.
A more insightful output, I think, could be this:

    java.lang.AssertionError: 
      Expected: a collection with size a value greater than <3>
        but: collection size <2> was less than <3>
        Actual collection:
          - 57
          - 965

By knowing what the array looked like at run time, it's more likely we'll be able to figure out what went wrong by just reading the error message.

This applies to any matcher dealing with the size of some data structure: hasSize, iterableWithSize, arrayWithSize.

In case the actual collection is very large, we should truncate it, as we don't want to pollute the output. We may save the full output to a text file for later consumption by the developer/tester:

    java.lang.AssertionError: 
      Expected: a collection with size a value greater than <9000>
        but: collection size <8999> was less than <9000>
        Actual collection:
          - 57
          - 965
          - 4394
          - ... (truncated; please check the full content at /tmp/hamcrest/$timestamp)

PR #294, opened 2020-04-19
Size matchers to provide insightful debugging information

alb-i986:
Resolving my own #174.

I'm leaving the commits separate for easier reviewing, I'll squash them afterwards.

Briefly, the solution introduces a base class (I called it `SizeMatcher`) for the matchers which match on the size of a collection-like object.
Although, I have to say, I'm not a huge fan of complex hierarchies (low coupling, you know), but it made perfect sense to align with the current design. Maybe that could be a separate topic to discuss, if you think it's something which could be discussed.

One thing to note is the naive algorithm I'm using to generate the correct indefinite article "a"/"an", in `SizeMatcher#indefiniteArticle`.
I guess you must have already bumped into this problem, so if you already have a method to solve it more elegantly, I'll be more than happy to use it.

Please check the commit messages for more details on the changes (I had to update a few tests).

Thanks


sf105:
This is an interesting idea. The reason we didn't show the actual collection contents originally is that we had no idea how big it would be (could be very large in some cases), so we left it out. Also, where there are compound matches on a collection, each one might dump out the whole contents.

I can see your point, but I'm not convinced that the removed duplication is worth the extra code. My experience is that the flakiness of the Java APIs sometimes justifies some clumsy code.

alb-i986:
> This is an interesting idea. The reason we didn't show the actual collection contents originally is that we had no idea how big it would be (could be very large in some cases), so we left it out. Also, where there are compound matches on a collection, each one might dump out the whole contents.

Cool, so if you agree that it is a concern, shall we add a truncation logic as I was suggesting in #174 ?

> I can see your point, but I'm not convinced that the removed duplication is worth the extra code. My experience is that the flakiness of the Java APIs sometimes justifies some clumsy code.

You mean the SizeMatcher? You prefer duplicating the code in every IsXXXWithSize class?
Between adding a layer to a complex hierarchy and duplicating, I prefer the former, in this case.
Because
1. we're talking about duplicating code in 4 classes, not 1 or 2
2. It just fits perfectly in the current design
3. I said "complex hierarchy" but it was not _that_ complex before my change, and it isn't after. It's still pretty much readable, understandable. Not the worst hierarchy I've seen.

My opinion is that if you want to re-think the design, I'm up for it. But until then, let's stick with what is already in place.

nhojpatrick:
@alb-i986 looks good, i've created a custom matcher myself, so the output reports what is in the collection is it is the wrong size. please can you rebase from master, as `hamcrest-core` and `hamcrest-library` have been refactored a lot and also deprecated, so that everything is just in `hamcrest`.

nhojpatrick:

PR #296, opened 2020-04-19
Fix feature matcher when feature throws

alb-i986:
`FeatureMatcher#featureValueOf` may throw.
Right now, that may happen in:
- HasToString, if the actual object has a custom toString method which throws
- FileMatchers.aFileWithCanonicalPath and siblings

I'm gonna squash as soon as you review.


sf105:
Thanks for contributing. I disagree with this one because if there's an exception at this level, then it's not a mismatch, but something is deeply broken and we should stop immediately.

alb-i986:
Well, for the toString method, you have a point. But we can't say that in general.
Look at [FileMatchers#aFileWithCanonicalPath](https://github.com/hamcrest/JavaHamcrest/blob/5dd6a622c73daf716c6e8f33efa712be8dbb42ed/hamcrest/src/main/java/org/hamcrest/io/FileMatchers.java#L55
): actualFile.getCanonicalPath() may throw IOException, which totally depends on the actual value. There's nothing deeply broken here.

sf105:
I would argue that that is different, because it reports a known possible failure that is in the domain of the matcher. For generic feature matchers, it's harder to make universal decisions. If you wanted to enforce consistency, I'd rather remove the exception handling from the file matcher.

What does the additional complexity handling of the exception do to improve the user's experience?

alb-i986:
> I would argue that that is different, because it reports a known possible failure that is in the domain of the matcher.

This is really where we disagree. I believe it's a "user space" issue, instead. Not a Matcher problem.

It's the same concept as throwing in matchesSafely() if the method for matching throws.
We don't do that, of course. We try-catch and return false. See e.g. [HasProperty](https://github.com/hamcrest/JavaHamcrest/blob/c1f7ab1adcd2ff8e268a2ef715d619599e181127/hamcrest/src/main/java/org/hamcrest/beans/HasProperty.java#L26)

nhojpatrick:
@alb-i986 please can you rebase from master, as `hamcrest-core` and `hamcrest-library` have been refactored a lot and also deprecated, so that everything is just in `hamcrest`.

nhojpatrick:

PR #303, opened 2020-05-06
Update Travis CI syntax

edumco:
### This pull request brings the following changes:

- Adds explicit `os` definition: Github now can build using macos and windows so it is a good practice include the `os` definition on `yml`.

- Re-order commands to follow Travis order of execution.

- Adds new java versions `openjdk11` and `openjdk14`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,15 @@
 language: java
 
+os: linux
+
 # Need to select an older Ubuntu distribution that supports JDK7
 dist: trusty
+
+jdk:
+  - openjdk7
+  - oraclejdk8
+  - openjdk8
+  - oraclejdk9
 
 # Workaround to using openjdk7 with Gradle due to security issue:
 # https://github.com/gradle/gradle/issues/2421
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,8 +1,18 @@
...
 # Need to select an older Ubuntu distribution that supports JDK7
 dist: trusty
 
+jdk:
+  - openjdk7
+  - oraclejdk8
+  - openjdk8
+  - oraclejdk9
+  - openjdk11
+  - openjdk14
```
nhojpatrick, line 14: 1) is early access of openjdk15 available yet?
2) is oraclejdk11 avaliable?

nhojpatrick:
Looks good, never tried travis myself as using circle ci or jenkins, but as it's showing green i'll approve as i assume that means all builds are passing

.travis.yml
edumco: I didnt try these versions but Travis has an option for allowing specific builds to fail without fail the complete build. Some projects adds the latest or beta versions in this list.

If you want i can create another PR to add these and other versions :dancers:

PR #306, opened 2020-06-12
expose hasLength(Matcher) in Matchers

chabala:
`hasLength(Matcher)` signature already exists, but currently client code must import `CharSequenceLength` directly in order to use it, unlike `hasLength(int)` which is already present.

This PR exposes it in the main Matchers object, similarly to how `hasSize(Matcher)` and `hasSize(int)`  are both present.

### DIFF
--- a/hamcrest/src/main/java/org/hamcrest/Matchers.java
+++ b/hamcrest/src/main/java/org/hamcrest/Matchers.java
@@ -1507,6 +1507,21 @@ public class Matchers {
     return org.hamcrest.text.StringContainsInOrder.stringContainsInOrder(substrings);
   }
 
+  /**
+   * Creates a matcher of {@link CharSequence} that matches when a char sequence has the length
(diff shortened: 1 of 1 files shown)


chabala:
Original CharSequenceLength PR #179

nhojpatrick:
@chabala please can you rebase from master, as `hamcrest-core` and `hamcrest-library` have been refactored a lot and also deprecated, so that everything is just in `hamcrest`.

chabala:
The change is not in `hamcrest-core` nor `hamcrest-library`. Do you still need a rebase @nhojpatrick? There have only been three commits since my branch 2 weeks ago.

chabala:
@nhojpatrick Are you waiting on me?

nhojpatrick:
@chabala personally I hate `org.hamcrest.Matchers` being a monolith of method calls. But as it still exists it would make sense to have methods people want moved into it.

PR #317, opened 2020-09-05
fix: contains/containsInAnyOrder overloading issue

### LINKED ISSUE #301: Casted Type should not matter for containsInAnyOrder
Regard following Example:

    @Test
    public void hamcrestTest(){
        String[] truth = new String[] {"1","2"};
        assertThat(Arrays.asList("1","2"),Matchers.containsInAnyOrder(truth));
        Object otherTruth = truth;
        assertThat(Arrays.asList("1","2"),Matchers.containsInAnyOrder(otherTruth));
    }

Result:
```
java.lang.AssertionError: 
Expected: iterable over [["1", "2"]] in any order
     but: Not matched: "1"

	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:956)
	at org.junit.Assert.assertThat(Assert.java:923)
	at or.example.Example.hamcrestTest(Example.java:22)
```

The first assert is successfull the second isn't. 
This leads to very subtle bugs.

Version: hamcrest-core:1.3

Actual Behavior:
If an array is casted to an object, containsInAnyOrder (and most probably others) regards it as a single item. Thus it asserts that the array ["1","2"] is in the results.

Expected Behavior:
Both calls should behave the same.

gokareless:
Fixes #301. 

**Caused:**
- Matcher instantiation depends of `actual` object type

**Fix:**
- Adjusted API to work with arrays embedded in objects

The assumption is that whenever we get to the `contains(T... items)` with an array, the only possible way to get there with an array is by passing `Object` type reference (because of `contains` overloaded methods). Therefore we can safely add extra-check for embedded array

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,6 +3,10 @@ name: Java CI
 
 on: [push]
 
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     name: Build Default Java/OS
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,15 +4,30 @@ name: Java CI
 on: [push]
 
 jobs:
+  build:
+    name: Build Default Java/OS
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK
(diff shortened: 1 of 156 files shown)


gokareless:
Adjusted in #380 as per @nhojpatrick request

PR #318, opened 2020-10-01
Fix typo in Javadoc

perlun:


nhojpatrick:
cheers for the patch, and cheers for caring about hamcrest

i still can't perform releases so not sure when your change will be included in a release

perlun:
Thanks @nhojpatrick. Doesn't sound too great. Have the previous maintainer(s) gone MIA?

PR #319, opened 2020-10-19
Added github action to build on each push

HanSolo:
Added a github action to build on each push. The build will run on Ubuntu 18.04 using jdk 7, 8, 9, 11, 14 and 15


nhojpatrick:
cheers @HanSolo keep meaning to look into github actions but never get around to it, will get this merged and see how it works out

Issue #325, opened 2021-01-27
IsBlankString matcher is inconsistent with String.isBlank()


TomRK1089:
The naming of the matcher implies it would give consistent results with `String.isBlank()`. However, it does not. The matcher uses a regex with pattern `\\s*` which will only match ASCII whitespace characters. `String.isBlank()` eventually delegates to `Character.isWhitespace()` which checks for Unicode whitespace as well. See https://stackoverflow.com/a/57998178 for a detailed breakdown, summarized below:

> But further, there are different definitions of white-space
> 
> Character.isWhitespace:
> 
> Determines if the specified character (Unicode code point) is white space according to Java. A character is a Java whitespace character if and only if it satisfies one of the following criteria:
> 
> It is a Unicode space character (SPACE_SEPARATOR, LINE_SEPARATOR, or PARAGRAPH_SEPARATOR) but is not also a non-breaking space ('\u00A0', '\u2007', '\u202F').
> It is '\t', U+0009 HORIZONTAL TABULATION.
> It is '\n', U+000A LINE FEED.
> It is '\u000B', U+000B VERTICAL TABULATION.
> It is '\f', U+000C FORM FEED.
> It is '\r', U+000D CARRIAGE RETURN.
> It is '\u001C', U+001C FILE SEPARATOR.
> It is '\u001D', U+001D GROUP SEPARATOR.
> It is '\u001E', U+001E RECORD SEPARATOR.
> It is '\u001F', U+001F UNIT SEPARATOR.
> the \s pattern (by default):
> 
> \s A whitespace character: [ \t\n\x0B\f\r]

I discovered this when using JUnit Quickcheck with an assumption to prevent handling whitespace strings of the form:
`Assume.assumeThat(string, not(blankString()));`
The assumption passed, and my code would check the string via `String.isBlank()` and fail the test.

hakanai:
For those of us with more experience it doesn't imply anything at all, because "blank" is such an ambiguous concept that any method in a library claiming to detect "blankness" is immediately banned for use by anyone who cares about what they're letting into their project.

TomRK1089:
In fact, "blank" does have a precise meaning in the Java ecosystem as I pointed out in my original post -- it should be consistent with the `isBlank` and `isWhitespace` methods of `String` and `Character` respectively. 

[...]

PR #326, opened 2021-02-05
Make IsBlank matcher consistent with String.isBlank

TomRK1089:
Fixes https://github.com/hamcrest/JavaHamcrest/issues/325

### DIFF
--- a/hamcrest/src/main/java/org/hamcrest/text/IsBlankString.java
+++ b/hamcrest/src/main/java/org/hamcrest/text/IsBlankString.java
@@ -5,8 +5,6 @@ import org.hamcrest.Description;
 import org.hamcrest.Matcher;
 import org.hamcrest.TypeSafeMatcher;
 
-import java.util.regex.Pattern;
-
 import static org.hamcrest.core.AnyOf.anyOf;
 import static org.hamcrest.core.IsNull.nullValue;
 
@@ -18,13 +16,11 @@ public final class IsBlankString extends TypeSafeMatcher<String> {
     @SuppressWarnings("unchecked")
     private static final Matcher<String> NULL_OR_BLANK_INSTANCE = anyOf(nullValue(), BLANK_INSTANCE);
 
-    private static final Pattern REGEX_WHITESPACE = Pattern.compile("\\s*");
-
     private IsBlankString() { }
 
     @Override
     public boolean matchesSafely(String item) {
-        return REGEX_WHITESPACE.matcher(item).matches();
+        return item.isBlank();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


TomRK1089:
Well...from Travis I learned that Hamcrest still supports Java7, hence why isBlank is not being used. Need to consider a different approach.

hamcrest/src/main/java/org/hamcrest/text/IsBlankString.java
```diff
@@ -20,7 +20,16 @@ private IsBlankString() { }
 
     @Override
     public boolean matchesSafely(String item) {
-        return item.isBlank();
+        final int length = item.length();
+        int offset = 0;
+        while(offset < length) {
```
peterdemaeyer, line 25: Consider sticking to the Hamcrest code style, which has a space before `if` and opening parenthesis.
It's not enforced in the build apparently, but still...

hamcrest/src/main/java/org/hamcrest/text/IsBlankString.java
```diff
@@ -20,7 +20,16 @@ private IsBlankString() { }
...
+            final int codePoint = item.codePointAt(offset);
+            if(!Character.isWhitespace(codePoint)) {
```
peterdemaeyer, line 27: Same as above: consider sticking a space between `if` and the opening parenthesis.

hamcrest/src/main/java/org/hamcrest/text/IsBlankString.java
TomRK1089: Can do. Are you using an autoformatter like spotless? It has Maven/Gradle plugins. Might be worth pulling into a separate PR :)

TomRK1089:
@peterdemaeyer looks like the builds have passed again, and the whitespace changes have been made :)

hamcrest/src/main/java/org/hamcrest/text/IsBlankString.java
peterdemaeyer: Thanks for fixing.
I was talking about a checker rather than a formatter.
JavaHamcrest is already using CheckStyle, but apparently it doesn't check much besides the fact that there are no tabs.
For the record: I'm not an official contributor to JavaHamcrest, I'm just a random someone with a stake in the project like you, I don't have the power to merge this PR even if I wanted to.

I imagine you inspired your implementation on Java 11 JDK's sources?
They look similar, yet a little bit different, and I wonder why...

The JDK's implementation distinguishes between Latin-1 and UTF-16 encoded strings. From what I can tell a `String` can be encoded in either of these 2. Your implementation only reflects the JDK's UTF-16 implementation. Is it safe to assume UTF-16? Does that also cover Latin-1? If such an assumption were safe, why does the JDK's implementation bother to make a difference?

In principle, I agree with this change, I'm just wondering if we should hold off fixing until JavaHamcrest bumps source/target compatibility to Java 11, in which case the fix will become simpler (i.e. your first commit).

Issue #323, opened 2021-01-04
Make hasXPath apply valueMatcher against multiple elements


fabian-froehlich:
It seems convinient to have a XPath-Matcher that is not limited to only look into the first matched element for the XPath

Example-XML:
```
<books>
	<book>
		<isbn>0-2556-3221-5</isbn>
		<name>Book-A</name>
	</book>
	<book>
		<isbn>0-5048-5640-5</isbn>
		<name>Book-B</name>
	</book>
</books>
```
will not be accpeted by the assert
`assertThat(xml, hasXPath("/books/book/isbn", is(equalTo("B-ISBN"))));`
```
Expected: an XML document with XPath /books/book/isbn is "B-ISBN"
     but: mismatches were: [an XML document with XPath /books/book/isbn is "B-ISBN" was "A-ISBN"
```
But this should match because the book with `B-ISBN` exists.
	 
A workaround would be to leave out the `valueMatcher` and put the value into the XPath
`assertThat(actualDocument, hasXPath("/books/book/isbn[text()='B-ISBN']"));`
which will correctly match the existence but at costs of a poorly explaining error message. This makes it harder identify whether your test is broken or your system-under-test.

night-gale:
Hi, I am currently working on this issue @nhojpatrick , and had a PR

PR #337, opened 2021-04-25
fixed Matcher.hasXpath against multiple element

night-gale:
issue #323 mentioned that `Matcher.hasXpath` could have unexpected behavior when given XML document with elements of the same path. The reason is that current logic only evaluates a single element out of an XML doc. It is now fixed by letting the matcher first get all possible elements. Then if any of them matches, the result will be a match, otherwise failed.


nhojpatrick:

PR #339, opened 2021-04-25
Added equalToWithDiff for comparing strings with diff message

### LINKED ISSUE #209: `equalTo` matcher should show a diff of the difference to make it possible to see where it is
If large texts are compared with `equalTo`, you just get both texts in full, and ultimately have to copy them out to text files manually and run a diff tool to figure out what the difference is. This is made worse because downstream tools reading the results don't produce diffs themselves. (To point the finger at one of the major suspects, IntelliJ IDEA shows diffs but only for tiny strings, even though for tiny strings you don't really need a diff to see the difference, but then doesn't show it for large strings, where you certainly can't see the difference without using diff.)

I think it's reasonable for the matcher's description of the mismatch to include this sort of thing as the primary message, and maybe include the full messages only for further information.

night-gale:
Fixed #209 
Added a new Matcher `equalToWithDiff` for compare string. When mismatch happens, instead of output two strings, only lines that mismatch would be output. For words in those lines that mismatch would be surrounded by `~`, the expected words would be surrounded by `**`.
* To implement this, the external library `io.github.java-diff-utils:java-diff-utils:4.5` is introduced.
* test class `IsEqualWithDiffTest` is created

### DIFF
--- a/hamcrest/hamcrest.gradle
+++ b/hamcrest/hamcrest.gradle
@@ -6,6 +6,7 @@ dependencies {
     testImplementation(group: 'junit', name: 'junit', version: '4.13') {
         transitive = false
     }
+    implementation "io.github.java-diff-utils:java-diff-utils:4.5"
 }
 
 jar {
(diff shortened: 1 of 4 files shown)


night-gale:
The library `io.github.java-diff-utils:java-diff-utils:4.5`  uses some jdk1.8 feature. That could be the reason why the compilation with jdk1.7 fails.

nhojpatrick:

PR #366, opened 2022-02-13
GitHub actions setup-java@v2

nhojpatrick:


nhojpatrick:
Looks like it requires a newer gradle version.

PR #383, opened 2022-03-17
Implement a matcher for parallel runs in iterables

Caster:
Introduce a new matcher `containsParallelRunsOf` and express the existing `containsInRelativeOrder` matcher using the new matcher with a single run as argument.

### Use case
In a project I am working on, we have a piece of code that logs a couple of messages and is being called multiple times in parallel. In the unit test of that code, we want to verify that the logs are correct. This is not straightforward, because the logs can be mixed up because of the parallel calls. Still, we want to check that (1) all logs are there and (2) their relative order is correct.

This is essentially a slightly more general version of the existing `containsInRelativeOrder` matcher, which is why I was able to express that matcher using the new one in this MR.


rmcdouga:

Caster:
Thank you for your efforts, @rmcdouga! I'll close this PR and watch yours. Hope we will see a new release.

PR #386, opened 2022-04-30
Typo fix

hextriclosan:
Typo fix in the comments


hextriclosan:
@nhojpatrick take a look, please

rjmunro:
When I see "iff" in a comment, I assume it means "if and only if" https://en.wikipedia.org/wiki/If_and_only_if

PR #414, opened 2024-07-23
Fix licence and add correct SPDX licence identifier

### LINKED ISSUE #264: Please add LICENSE as META-INF/LICENSE, add Bundle-License header
1) Thanks for Hamcrest

2) Current hamcrest.jar are hard to bundle into other software due to missing licensing information.

Suggested changes are:

2.1) Package license as `META-INF/LICENSE`. Current one is `LICENSE.txt`, and de-facto standard is `META-INF/LICENSE`. The license files packaged at the root cause problems with Android.
Basically whoever makes an uberjar they have to implement "license merger", and typically they configure for `META-INF/LICENSE`.
2.2) Please add `Bundle-License: BSD-3-Clause` manifest entry (where `BSD-3-Clause` is SPDX identifier, see https://osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module-bundle-license )
Unfortunately, there are lots of BSD-style licenses, and it is hard to tell what "New BSD License" means.
2.3) Even though 2.1 and 2.2 would be great, you might want to use BSD-3-Clause for `<licence><name>` tag in `pom.xml`. The use of SPDX in pom is not common yet, however BSD-3-Clause there would help even humans.

### LINKED ISSUE #355: Conflicting license declarations
Hi there,

While checking open source licenses used by my application's dependencies, I noticed Hamcrest 1.3 was published with an odd license declaration, "[New BSD License](http://www.opensource.org/licenses/bsd-license.php)". This should have been "2-Clause BSD License", but it is published now.

It looks like build.gradle has been updated to "[BSD License 3](http://opensource.org/licenses/BSD-3-Clause)" but the LICENSES.txt in the repository is still the "2-Clause BSD License".

LICENSES.txt should contain the 3-Clause BSD License and build.gradle should reference "3-Clause BSD License" as the license name. Our Java plugin for open source license reporting now works around these issues, but it might be convenient for other people inspecting their licenses.

tumbarumba:
* Moved `LICENSE.txt` to the more standard `LICENCE`, ensuring the text is the same as the canonical version at https://opensource.org/license/BSD-3-Clause
* Used the correct SPDX license identifier in the Bundle-License field in the manifest, as well as the pom
* Used the correct SPDX license identifier in the pom

Closes #264, Closes #355, Closes #399


vlsi:
Thanks for looking into this.

I'm afraid the PR does not fix the core of #264.

See the second clause of the license:

> 2. Redistributions in binary form must reproduce the above copyright notice

Imagine somebody redistributes `hamcreset.jar` or includes it in their application. The license forces them to include the copyright notice, however, the release jars on Maven Central do not include the full license text. How would they know which exact license text (==copyright notice) matches every Hamcrest release?

See https://github.com/hamcrest/JavaHamcrest/pull/399/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R75-R77

The link https://opensource.org/license/BSD-3-Clause points to a license **template** rather than a canonical text. Please make the link point to the license text, and and please include the license text as `META-INF/LICENSE` to make it easier for the consumers to comply with the license.

tumbarumba:
Thanks for looking at this @vlsi. I'll make the suggested changes in the next few days

tumbarumba:
Fix merged. @vlsi, can you please check?

Issue #252, opened 2019-03-22
Generic type bounds of IsIterableContaining are inconsistent with those of IsMapContaining


nibsi:
Hello all! Thanks a lot for the new Hamcrest 2.1 library, it's a pleasure to work with and I can also use it easily in my Java 9 modules.

The `IsIterableContaining` matcher extends `TypeSafeDiagnosingMatcher<Iterable<? super T>>`. I understand that the goal is to, for instance, match a `Set<Number>` when it has an `Integer`.

Why then does `IsMapContaining` extend `TypeSafeMatcher<Map<? extends K, ? extends V>>`? Wouldn't it be just as reasonable to match a `Map<Object, Object>` when it has a `String` key? Or a `Long` value?

tumbarumba:
Writing library code with Java generics is a bit like (╯ರ ~ ರ）╯︵ ┻━┻

Perhaps @sf105 might have some insights as to how the current implementation came to be the way it is?

Nibsi:
```java
Collection<Banana> bananas = singleton(banana);
Collection<Fruit>  fruits  = singleton(banana);
    
assertThat(fruits,  hasItem(banana));
assertThat(bananas, hasItem(fruit) );  // this one is currently invalid
```
I think this makes sense: Why would I want to check if a collection of some specific type contains some less specific type? Usually I already know the actual runtime type of the value I use to create the matcher with, because I create it myself, whereas the collection being examined is passed in and could be anything.

The same is true for maps. Therefore I would propose to make the `hasKey()` and `hasValue()` matchers contravariant.

foal:
> The IsIterableContaining matcher extends `TypeSafeDiagnosingMatcher<Iterable<? super T>>`. 

The IsIterableContaining  have to extend `TypeSafeDiagnosingMatcher<Iterable<? extends T>>`.  
Usualy you have a collection of entities (e.g. Car). This list can contain any object that extends Car (`List<? extent Car>`). And you want to apply to each element the Matcher that can check the Car or it's parent (`Matcher<? super Car>`).

So, you want to construct Matcher that take the Iterable<? extends Car> and apply to it given Matcher<? super Car>

tumbarumba:
[...]

PR #422, opened 2024-09-08
Allow matching against polymorphic collections

tumbarumba:
As discussed in #252, there are problems with matching against polymorphic collections.

This fix attempts to apply the PECS rule (producer extends, consumer super) to the Hamcrest IsIterableContaining matcher. In this instance, a collection of items should be treated as a producer according to this rule, while a matcher acts as a consumer. There was some confustion about PECS in the context of collections of matchers, but I'm hoping this change addresses those issues.

I'm keen for others to test this work out

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,22 @@
 
 ## Version 3.1 (Unreleased)
 
+### Breaking Changes
+
+* As a result of the bugfix to allow matching against polymorphic collections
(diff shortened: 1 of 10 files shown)


foal:
Nice! 
About PECS, INHO it is clear - the `itemMatcher` is consumer - it consumes the elements of the target collection to check them.It may check the element or any **parent** of the element (e.g. object != null). The output is producer - it produces global matcher for a whole collection and should be applied to collection of the elements or to collection of element's **children**.

PR #433, opened 2025-04-01
Close code section correctly

joegallo:
to fix javadoc formatting for `notNullValue`.

I noticed that https://hamcrest.org/JavaHamcrest/javadoc/3.0/org/hamcrest/Matchers.html appears to be a bit broken, and it seems like if this `<code>` section were terminated earlier then things would work correctly.


rmcdouga:

